PR #3, opened 2024-04-10
[DPE-3964;DPE-4044] spark-test package alpha version

deusebio:
This PR creates a first skeleton and functionalities to be used when testing Charmed Spark. In particular, it provides fixtures for creating and managing the different resources, such as:

* K8s namespaces
* spark8t Service Accounts
* Pods with the Charmed Spark
* S3 Buckets

More information in the readme associated. 

An example of how fixtures can be used are available in the integration tests.


python/tests/integration/test_s3.py
```diff
@@ -0,0 +1,77 @@
...
+def test_spark_iceberg_integration(pod, service_account, registry, bucket, credentials):
+
+    num_rows = 4
+
+    bucket.upload_file(RESOURCES / "python" / "test-iceberg.py")
+
+    from spark8t.domain import PropertyFile
+
+    extra_confs = PropertyFile(
+        {
```
theoctober19th, line 30: I think it would be awesome if we can gather the commonly used spark properties somewhere as well (maybe a fixture?) so that these need not be typed in every test case.

python/spark_test/core/s3.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if cls._exists(bucket_name, s3):
+            raise ValueError(f"Cannot create bucket {bucket_name}. Already exists.")
+
+        s3.create_bucket(Bucket=bucket_name)
+
+        return Bucket(s3, bucket_name)
+
+    def init(self):
+        self.s3.put_object(Bucket=self.bucket_name, Key=("spark-events/"))
```
theoctober19th, line 45: question: Are we creating `spark-events` directory in the bucket by default?

python/spark_test/core/s3.py
```diff
@@ -0,0 +1,91 @@
...
+    access_key: str
+    secret_key: str
+    host: str
+
+    @property
+    def endpoint(self) -> str:
+        return f"http://{self.host}:80"
+
+
+class Bucket:
```
theoctober19th, line 20: We may have the need to create nested path structure (eg, `bucket/foo/bar/abc.txt`). A function that lets this path to be created could also be useful here.

python/spark_test/core/pod.py
```diff
@@ -0,0 +1,99 @@
+import subprocess
+from pathlib import Path
+from typing import Dict, List
+
+from spark_test import BINS, PKG_DIR
+
+
+class Pod:
```
theoctober19th, line 8: thought: One thing that we can enhance in this framework starting from the beginning is the logging mechanism. This would save us a lot of time and frustration during debugging.

theoctober19th (approved):
I reviewed the PR and it looks really nice and provides a good foundation for us to start migrating tests from existing repos to this new repo. There are a few typos here and there, but I have not pointed them out at the moment because I reviewed the architecture and concept and not the nitpicking details.

welpaolo (approved):
Looks a really good start :) 
I left some small comments and suggestions.

Some docstrings are missing but I believe we can improve this in the next iterations.

python/tests/integration/test_s3.py
deusebio: Yes, totally! +1

python/spark_test/core/s3.py
deusebio: I believe this would be required for storing the logs, for the history server. I have copied this from elsewhere, maybe when testing we can check if this is really necessary

python/spark_test/core/s3.py
deusebio: agreed, but can we do this later when needed? I was somewhat sticking to what we had already as business-logic in the other parts. I would add features to these as we go, based on need if that makes sense?

python/spark_test/core/pod.py
deusebio: Totally! I really like this! I'll do this straight away, as I need this already in the tests!

deusebio:
I have added a bunch of documentation and addressed the points raised. In particular, I have added the ability to retrieve logs in the class.

python/spark_test/core/s3.py
theoctober19th: Yes, no problem.

python/spark_test/core/pod.py
theoctober19th: @welpaolo What I mean is adding debug logs at places so we have an idea of where the code failed. Your suggestion is worth looking at, too.

theoctober19th (approved):

welpaolo (approved):
LGTM

PR #5, opened 2024-04-23
[DPE-3966] Spark Bundle with YAML

deusebio:
Adding the bundle using YAML + some integration testing. 

I have laid out a structure in the integration testing for switching between a basic bundle (with the basic spark component) to the version with observability integrated. When integrating with observability, we need both to deploy more charms (e.g. pushgateway, agent, etc) but also deploy the cos bundle on a separate model. 

The cos bundle in a separate model has been done as a fixture such that we could also use it in other charms and be able to perform the same test with and without observability. 

The switch can simply be done using an extra argument (or a the cos_model fixture). If this variable is None, cos (and related charms) will not be deployed. When it is valued, we check whether a model named in the same way already exists. If it exists, the cos bundle is not deployed. If it does not exist, it gets deployed. 

**NOTE** Unfortunately, although the deployment with COS works properly on my machine, in CI it keeps failing. I'm trying to figure out why, but I will definitely benefit from a first round of review. 

**NOTE 2** Trying to fix the CI with COS, I have also realized that there were some errors raised (due to the fact that relation exists, but databag is not yet populated, that typically occurs when deploying all at once in bundles). I have tried to slightly improve this in two PRs for Kyuubi and Integration-hub:
* https://github.com/canonical/kyuubi-k8s-operator/pull/11
* https://github.com/canonical/spark-integration-hub-k8s-operator/pull/6
* https://github.com/canonical/spark-history-server-k8s-operator/pull/40


releases/3.4/bundle.yaml.j2
```diff
@@ -0,0 +1,74 @@
...
+    channel: 3.4/stable
+    revision: 15
+    resources:
+      spark-history-server-image: 9
+    scale: 1
+    constraints: arch=amd64
+  integration-hub:
+    charm: spark-integration-hub-k8s
+    channel: latest/edge
+    revision: 3
```
deusebio, line 58: Revision based on https://github.com/canonical/spark-integration-hub-k8s-operator/pull/6

releases/3.4/bundle.yaml.j2
```diff
@@ -0,0 +1,74 @@
+bundle: kubernetes
+applications:
+  kyuubi:
+    charm: kyuubi-k8s
+    channel: "latest/edge"
+    revision: 5
```
deusebio, line 6: Revision based on https://github.com/canonical/kyuubi-k8s-operator/pull/11

welpaolo (approved):
LGTM

theoctober19th (approved):
LGTM

releases/3.4/bundle.yaml.j2
```diff
@@ -0,0 +1,74 @@
...
+    scale: 1
+    options:
+      bucket: {{ bucket }}
+      endpoint: {{ s3_endpoint }}
+      path: spark-events
+    constraints: arch=amd64
+  history-server:
+    charm: spark-history-server-k8s
+    channel: 3.4/edge
+    revision: 24
```
deusebio, line 50: **TODO** Update with released version

releases/3.4/bundle.yaml.j2
```diff
@@ -47,7 +47,7 @@ applications:
   history-server:
     charm: spark-history-server-k8s
     channel: 3.4/edge
-    revision: 24
+    revision: 25
```
deusebio, line 50: from [here](https://github.com/canonical/spark-history-server-k8s-operator/releases/tag/rev25)

PR #7, opened 2024-04-29
[DPE-4229] Terraform bundle

deusebio:
Adding bundles based on Terraform. I have also added the tests to deploy it with and without cos. 

I was scouting a bit to see if it existed a Python library for binding with terraform, and I found these:

* https://pypi.org/project/python-terraform/
* https://github.com/Prodesire/py-libterraform

However, they seemed a bit out of date and not supporting terraform 1.8.x which I've been using. Earlier today, Pedro suggested to look into [this one](https://github.com/GoogleCloudPlatform/terraform-python-testing-helper), that should work nicely with pytest. However I had already implemented most of the stuff and I'll explore its use in a separate ticket. 

I'll move the PR into ready one the PR about the YAML will be merged #5


deusebio:
@marcoppenheimer

welpaolo (approved):
Looks good to me!

Copyrights are missing in several files.
I am not really familiar with terraform but the rest of the code looks fine with me.
Maybe it will be a good idea to ask @marcoppenheimer to review it.

releases/3.4/terraform/cos/applications.tf
```diff
@@ -0,0 +1,88 @@
...
+
+resource "juju_application" "cos_configuration" {
+  name = "cos-configuration"
+
+  model      = var.model
+
+  charm {
+    name    = "cos-configuration-k8s"
+    channel = "latest/stable"
+    revision = 46
```
theoctober19th, line 30: thought: I have a feeling that at one point, it would be too much hassle / complicated to track and change these revision numbers that are scattered around the files (in bundle yaml as well as terraform files). Ideally it would be awesome if we could track these centrally and have one place where this could be controlled from, however I don't have a concrete idea yet.

theoctober19th (approved):
I'm not much knowledgeable in Terraform, however to my eyes the request looks good. I have a few minor thoughts laid out.

releases/3.4/terraform/cos/applications.tf
deusebio: yes, I totally agree. I would like to get a bit more opinionated on the config for the TF, but I would prefer to keep it in a separate PR where we design the config and implement it properly

PR #18, opened 2024-05-14
[DPE-4121] Improvements on Spark E2E tests

theoctober19th:
This PR aims to:
1. Add a few integration tests related to Kyuubi bundle
2. Few minor improvements 
3. Add docstrings

### DIFF
--- a/python/tests/integration/conftest.py
+++ b/python/tests/integration/conftest.py
@@ -15,7 +15,17 @@ from pytest_operator.plugin import OpsTest
 
 from tests import RELEASE_DIR
 
-from .helpers import COS_ALIAS, Bundle, deploy_bundle, local_tmp_folder
+from .helpers import (
+    COS_ALIAS,
+    Bundle,
+    deploy_bundle,
+    deploy_bundle_terraform,
+    deploy_bundle_yaml,
+    get_kyuubi_credentials,
+    get_postgresql_credentials,
+    local_tmp_folder,
+    set_s3_credentials,
+)
 from .terraform import Terraform
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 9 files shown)


deusebio (approved):
Great Job! LGTM!

python/tests/integration/conftest.py
```diff
@@ -169,3 +183,50 @@ def create_file(path: Path, response: requests.Response):
 
         if cos_model:
             await ops_test.forget_model(cos_model)
+
+
+@pytest.fixture(scope="module")
+async def kyuubi_bundle(
```
welpaolo, line 189: I would like this to be renamed to spark_bundle

welpaolo: @deusebio what do you think ?

python/tests/integration/test_bundle.py
```diff
@@ -54,103 +35,10 @@ def namespace(namespace_name):
...
-        apps=applications,
-        timeout=2500,
-        idle_period=30,
-        status="active",
-        raise_on_error=False,
-    )
-
-    for app in applications:
-        assert ops_test.model.applications[app].status == "active"
+async def test_deploy_bundle(ops_test, kyuubi_bundle):
```
welpaolo, line 40: This should be named `spark_bundle`

welpaolo (commented):
All good, just want a clarification for @deusebio about how to name the bundle

python/tests/integration/conftest.py
deusebio: Yes, that's a good idea!

theoctober19th: I agree. Just updated.

python/tests/integration/test_bundle.py
theoctober19th: Update.

welpaolo (approved):
LGTM

PR #21, opened 2024-05-23
[MISC] Charmed Spark K8s published bundle

deusebio:
This PR encode what was uploaded to charm hub to release the bundle. I feel it would be worth keeping this in the repo for reference


theoctober19th (changes requested):
The bundle is missing the Azure Storage Integrator charm and its relation (which exist in the `main` branch).

PR #26, opened 2024-06-28
[DPE-4807] Fix bundle tests

deusebio:
The tests updates both the versions of the workloads, but also - most importatnly - move the CI on self-hosted runners. 

In regular github runners, the deployment has now become a bit too large, making things very flaky. Especially when deployed alongside with COS, the tests often hung or failed, becuase of weird errors, that I (or others) could not reproduce locally or on larger machines ([reference](https://chat.canonical.com/canonical/pl/uqokjw66gpyxtem58rfcj657ro)). 

My thought here is that in those conditions, the containers becames either irresponsive or failing, because of concurrency to use the same limited resources.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -112,6 +112,7 @@ jobs:
             echo Skipping unstable tests
             echo "mark_expression=not unstable" >> $GITHUB_OUTPUT
           fi
+      # Commented debugging
       # - name: Setup upterm session
       #   uses: lhotari/action-upterm@v1
       - id: tests-integration
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -45,7 +45,6 @@ jobs:
 
   integration-tests:
     name: Integration Tests
-    # runs-on: ubuntu-22.04
     runs-on: [self-hosted, linux, X64, jammy, xlarge]
     timeout-minutes: 120
     strategy:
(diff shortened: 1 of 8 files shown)


python/spark_test/resources/bin/pod.sh
```diff
@@ -9,7 +9,7 @@ wait_for_pod() {
   POD=$1
```
theoctober19th, line 9: i was interested to see if we could use `kubectl wait` based waiting here, just like we have done it in rock and snap tests  (eg, [here](https://github.com/canonical/charmed-spark-rock/blob/3.4-22.04/edge/tests/integration/utils/k8s-utils.sh#L16) ) We maybe can even reuse the `k8s-utils.sh` file and its utility functions here as well.

theoctober19th (approved):

python/spark_test/resources/bin/pod.sh
deusebio: Oh nice, just testing this in https://github.com/canonical/spark-k8s-bundle/pull/28

.github/workflows/ci-tests.yaml
```diff
@@ -82,13 +82,15 @@ jobs:
...
+          bootstrap-options: " --agent-version 3.4.3"
           microk8s-group: snap_microk8s
           microk8s-addons: "rbac dns minio metallb:${{ env.IPADDR }}-${{ env.IPADDR }}"
       - id: cache-image
         name: Cache Spark Image Locally
         run: |
           # Download image for avoiding time out
-          IMAGE="ghcr.io/canonical/charmed-spark:3.4-22.04_edge"
+          IMAGE="ghcr.io/canonical/charmed-spark-kyuubi@sha256:9268d19a6eef91914e874734b320fab64908faf0f7adb8856be809bc60ecd1d0"
+          # IMAGE="ghcr.io/canonical/charmed-spark-kyuubi:3.4-22.04_edge"
```
welpaolo, line 93: Do you want to keep this commented line ?

python/tests/integration/helpers.py
```diff
@@ -172,7 +172,7 @@ def get_secret_data(namespace: str, service_account: str):
 
 async def deploy_bundle_yaml(
     bundle: Bundle,
-    service_account: ServiceAccount,
+    # service_account: ServiceAccount,
```
welpaolo, line 175: Remove comment

python/tests/integration/helpers.py
```diff
@@ -219,7 +219,7 @@ async def deploy_bundle_yaml(
 
 async def deploy_bundle_terraform(
     bundle: Terraform,
-    service_account: ServiceAccount,
+    # service_account: ServiceAccount,
```
welpaolo, line 222: Remove comment

welpaolo (approved):
LGTM, just left some minor comments

python/spark_test/resources/bin/pod.sh
deusebio: Uhm, it has more failure than this PR. I would leave it outside for now

PR #34, opened 2024-08-05
[DPE-5032] Add 3.5 release bundle

deusebio:


theoctober19th (approved):

.github/workflows/ci-tests.yaml
```diff
@@ -50,6 +50,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        version:
+          - --spark-version "3.5.1"
```
welpaolo, line 54: Do we want to run tests only on 3.5.X or also on 3.4.X ?

python/spark_test/fixtures/pod.py
```diff
@@ -20,20 +20,34 @@ def admin_pod_name():
 
 
 @pytest.fixture(scope="module")
-def admin_pod(kubeconfig, namespace, admin_pod_name):
+def spark_version():
+    return "3.4.2"
```
welpaolo, line 24: Here should be 3.5.1 ?

releases/3.5/terraform/cos/applications.tf
```diff
@@ -0,0 +1,91 @@
...
+
+  resources = {
+    git-sync-image = 32
+  }
+
+  config = {
+    git_branch = "main"
+    git_depth  = "1"
+    git_repo   = "https://github.com/canonical/spark-k8s-bundle"
+    grafana_dashboards_path = "releases/3.4/resources/grafana/"
```
welpaolo, line 44:
```suggestion
    grafana_dashboards_path = "releases/3.5/resources/grafana/"
```

welpaolo (approved):
LGTM, I left some comments

.github/workflows/ci-tests.yaml
deusebio: I believe so. But I wanted first to make sure that the tests were working on 3.4.2. Let me update this, but the flakiness of the tests is a bit problematic here.

python/spark_test/fixtures/pod.py
deusebio: This should not matter too much, as this is generally overridden by the flag provided in the input.

releases/3.5/terraform/cos/applications.tf
deusebio: Good catch!

deusebio:
As discussed, I'm going to merge, although tests are VERY flaky (failed tests needed to be re-run 6 times, to get them all to pass). We will address the flakyness in the [DPE-5078](https://warthogs.atlassian.net/browse/DPE-5078).

PR #36, opened 2024-08-06
[DPE-4845] Azure Storage integration with Spark K8s bundle  and end-to-end tests

theoctober19th:


python/spark_test/core/azure_storage.py
```diff
@@ -0,0 +1,128 @@
...
+    storage_account: str
+
+    @property
+    def connection_string(
+        self,
+    ) -> str:
+        return f"DefaultEndpointsProtocol=https;AccountName={self.storage_account};AccountKey={self.secret_key};EndpointSuffix=core.windows.net"
+
+
+class Container:
```
deusebio, line 25: **suggestion** Maybe we can inherit both s3 and azure from the same class, such that one can create functions/tests that are independent from the backend 

If not needed yet, I don't have strong feelings to either address this in this PR or not

python/tests/integration/conftest.py
```diff
@@ -81,6 +87,12 @@ def backend(request) -> None | str:
     return request.config.getoption("--backend")
 
 
+# @pytest.fixture(scope="module")
+# def object_storage_backend(request) -> None | str:
+#     """The object storage backend to be used in the bundle."""
+#     return request.config.getoption("--object-storage-backend")
```
deusebio, line 93: **thought** This would have been a good way to handle parametric backends, but if it is not implemented we can get rid of this

python/spark_test/core/azure_storage.py
```diff
@@ -0,0 +1,128 @@
...
+            container_client = self.client.get_container_client(
+                container=self.container_name
+            )
+            with open(file_name, mode="rb") as data:
+                container_client.upload_blob(blob_name, data=data, overwrite=True)
+        except Exception:
+            return False
+        return True
+
+    def list_blobs(self):
```
deusebio, line 119: **thought** maybe we can call this `list_objects` or simply `list`, with a signature similar from s3 and azure (in input and output)

deusebio (approved):
The PR looks good and sensible. I praise you for the good work, and I know that this hasn't been easy and required to address a number of other issues. I don't have any major concern with respect to the PR and the code added here. 

I only believe there is room for refactoring in order to reduce the copy and paste between tests. Right now the `*_azure` tests are very very similar one another. 

I checked and the differences seems to be related to:
* way of deploying => we could provide a standard API like deploy_bundle that inside handles the difference between the backends, using one file for S3 and another file for azure
* Bucket vs Container class and signature => here we could have the same API with two implementations (one for Bucket and one for Container). 
* File/scripts URI to be fed to spark-submit => also here we could have a general function to be used where we provide the two different implementations for the two backends

I feel it would not be massive work to refactor the code such that we only use one tests skeleton that can apply to both backends. 

Anyhow, this is more for getting a more maintainable code base, rather than having copy&paste of tests that are 90% similar. We can also address this as a tech dept in the next pulse, but if we agree on the principle, let's put a ticket in the backlog

theoctober19th:
@deusebio I agree with your point. Ideally, I see the tests parameterized based on the object storage backend. However, I wanted start with a simpler solution and keep the refactoring for a separate PR. I'll quickly create a Jira ticket for that. IMO all three of us in the team agree to this point.

theoctober19th:
https://warthogs.atlassian.net/browse/DPE-5065

python/tests/integration/conftest.py
theoctober19th: I started writing this initially but later commented because I want to address these improvements in a separate PR. I agree with your point.

python/spark_test/core/azure_storage.py
theoctober19th: I've created a new ticket and linked it in the comments below, I'll address all of these points in that PR.

PR #39, opened 2024-08-17
Vale test fixes

izmalk:
Fixing Vale tests for Charmed Spark documentation.


izmalk:
Why do we need both `index.md` and `overview.md`?

PR #42, opened 2024-09-02
[DPE-5078] Improve stability of Spark K8s bundle tests

theoctober19th:
image

The tests are now all green (on 2 consecutive runs). We may encounter some hiccups that were not seen ever before.. but the issues that were seen before have been solved.

### DIFF
--- a/python/spark_test/resources/bin/s3.sh
+++ b/python/spark_test/resources/bin/s3.sh
@@ -8,8 +8,15 @@ setup_bucket() {
   # errcho $(microk8s.enable minio)
 
   # Get Access key and secret key from MinIO
-  ACCESS_KEY=minio
-  SECRET_KEY=minio123
+  kubectl get secret -n minio-operator microk8s-user-1
+  if [ $? -eq 0 ]; then
+    ACCESS_KEY=$(kubectl get secret -n minio-operator microk8s-user-1 -o jsonpath='{.data.CONSOLE_ACCESS_KEY}' | base64 -d)
+    SECRET_KEY=$(kubectl get secret -n minio-operator microk8s-user-1 -o jsonpath='{.data.CONSOLE_SECRET_KEY}' | base64 -d)
+  else
+    echo "Using default MinIO access key and secret key..."
+    ACCESS_KEY=minio
+    SECRET_KEY=minio123
+  fi
 
   # Get S3 endpoint from MinIO
   S3_ENDPOINT=$(kubectl get service minio -n minio-operator -o jsonpath='{.spec.clusterIP}')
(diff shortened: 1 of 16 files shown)


python/spark_test/resources/bin/s3.sh
```diff
@@ -8,8 +8,8 @@ setup_bucket() {
   # errcho $(microk8s.enable minio)
 
   # Get Access key and secret key from MinIO
-  ACCESS_KEY=$(kubectl get secret -n minio-operator microk8s-user-1 -o jsonpath='{.data.CONSOLE_ACCESS_KEY}' | base64 -d)
-  SECRET_KEY=$(kubectl get secret -n minio-operator microk8s-user-1 -o jsonpath='{.data.CONSOLE_SECRET_KEY}' | base64 -d)
+  ACCESS_KEY=minio
+  SECRET_KEY=minio123
```
deusebio, line 12: **suggestion** if it is reasonably easy, here we could use the formulation that @welpaolo did that is compliant with both versions of the MinIO addon

python/tests/integration/helpers.py
```diff
@@ -122,17 +122,29 @@ def _local(pointer) -> str:
     )
 
 
-async def deploy_bundle(ops_test: OpsTest, bundle: Bundle) -> tuple[int, str, str]:
+async def set_memory_constraints(ops_test, model_name):
```
deusebio, line 125: **questions** what is this solving? I'm just thinking whether we should (not as part of this ticket though) bring this to other CIs in SPark (or even Kafka) as well

deusebio (approved):
AMAZING!!! Thanks @theoctober19th !

python/tests/integration/helpers.py
theoctober19th: The errors like `kyuubi-k8s` always stuck on "Waiting for Pebble" and `traefik-k8s` waiting for being idle had correlation with the error "Pebble socket not found, (is the container restarted?)" on the Juju debug logs. This log appeared very frequently in Juju logs and one of the suspected reason for this was that the pods were restarted very frequently and thats why the rescheduling of the pods were causing the Pebble layer to frequently disappear. One of the many reasons the pod reschedule could happen is due to OOM (Out of Memory)... which I found `mysql` team has similar observations and simlar fixes done [here](https://github.com/canonical/mysql-k8s-operator/issues/205) and [here](https://github.com/canonical/traefik-k8s-operator/issues/155).

python/spark_test/resources/bin/s3.sh
welpaolo: I also suggest that

welpaolo (approved):
LGTM! Great work here!

python/tests/integration/helpers.py
deusebio: This is really gold then! Well done on the triaging and fixing this issue! It was far from obvious!

PR #43, opened 2024-09-03
[DPE-5322] Update TF bundle with Azure storage integration 

deusebio:
The PR adds support in the main TF module for Azure storage integration. 

The PR provides a new TF variable (e.g. `storage_backend`) to switch between s3 or azure. 

If `storage_backend=azure` is provided, the deployment can be customized via the following object variable:

```
variable "azure" {
  description = "Azure Object storage information"
  type = object({
    container            = ...
    storage_account      = ...
  })
  default = {}
}
```

An example of a `tfvars.json` deployment for a Azure storage deployment would be:

```
{
    "storage_backend": "azure",
    "azure": {
	"storage_account": "charmedsparkstorage",
	"container": "dev"
    },
    "kyuubi_user": "kyuubi-engine",
    "model": "spark",
    "cos_model": "cos"
}
```

The deployment with Azure storage replaces the `s3-integrator` with the [`azure-storage-integrator`](https://charmhub.io/azure-storage-integrator), that provides a similar functionality of deliverying object storage credentials to the different charms. However Azure Storage Integrator improves the UX of `s3` by allowing the credentials to be fed via Juju secrets, provided via config options. 

In order to configure the `Azure Storage Integrator` charm, first create a secret holding the `secret_key`

```
juju add-secret azure-credentials secret-key=...
```

Next grant permissions:

```
juju grant-secret <secret_id> azure-storage
```

And finally configure the `Azure Storage Integration` charm 

```
juju config azure-storage credentials=secret:<secret_id> 
```

More information can be found in the [Charmed Spark Documentation](https://charmhub.io/spark-k8s-bundle) under the [How-to deploy](https://charmhub.io/spark-k8s-bundle/docs/h-deploy) and [How-to configure integration hub](https://charmhub.io/spark-k8s-bundle/docs/h-use-integration-hub) userguides.


releases/3.4/terraform/azure/applications.tf
```diff
@@ -0,0 +1,141 @@
...
+    name    = "azure-storage-integrator"
+    channel = "latest/edge"
+    revision = 1
+  }
+
+  config = {
+    container = var.azure.container
+    storage-account = var.azure.storage_account
+    path = "spark-events"
+    connection-protocol = "abfss"
```
theoctober19th, line 40: Should we probably have this from `var.azure` as well?

theoctober19th (commented):
I have only one concern regarding whether the bundle using Terraform and Azure backend is tested and verified to be working out of the box (because there were a lot of hurdles when doing so in YAML a few weeks).

theoctober19th (approved):

theoctober19th:
I found that Juju Terraform provider supports resource juju_secret as well ([ref](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/secret)). Currently, our deployment involves extra steps after doing `terraform apply ` (for exmaple, `juju add-secret`, `juju grant-secret`, `juju config...`) which is both an overhead as well as the entire deployment is unusable unless these extra steps are done.

I was thinking if it would be a good idea to get a secret key as Terraform var from the user during `terraform apply` and then use this value to create juju secret and grant it to the application on the fly, if that's possible (which on initial look seems possible to me).

That way, all the user needs to do in order to get of the bundle up and running is `terraform apply` (possibly with some vars). I also think when we fully support AKS deployment, we can have Terraform create the required resources like storage account, AKS clutser etc. and then supply the secret key of the storage account automatically during the provisioning.

What do you guys think about this?

deusebio:
@theoctober19th feels we are mind connected here and you are definitely SPOT ON! Just yesterday we were discussing about this. We are currently working with Diko (from ManSol) to test this on azure using their workflows, which yes, would eventually lead to provide everything. Right now we are just thinking of requiring the user to create their container, but the rest we should create it with the one-click deployment. 

However, I would like to keep our bundle/solution (deploying all Spark components on top of AKS) separated from the deployment (that deploys everything). Of course, the deployment would use the solution module.  

Anyhow, we actually have a few points that we discussed to refactor our modules before integrating everything. Let me gather all the facts, and then I'll write a bullet list of what we need to do :D

theoctober19th:
I'm removing myself from reviewer temporarily since it's always appearing on my list of PRs to review (which I see is not ready for review yet).

deusebio:
This is not relevant anymore, as it will be addressed in a subsequent PR

PR #49, opened 2024-10-06
[DPE-5526] Add monitoring checks to the bundle

welpaolo:
This PR checks in the bundle that the most recent releases of Charmed Spark components that include monitoring functionalities. 
This include metrics/logs for Spark History Server and Kyuubi.

In addition to provide more test stability I move back to Github runners since they can handle the deployment.


.github/workflows/ci-tests.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
 
   integration-tests:
     name: Integration Tests
-    runs-on: [self-hosted, linux, X64, jammy, xlarge]
```
deusebio, line 48: **question** Why don't need self-hosted runners anymore? Interesting...

deusebio (approved):
LGTM!

.github/workflows/ci-tests.yaml
welpaolo: The tests were failing on GitHub runners because of the lack of disk space. I just [free-up](https://github.com/canonical/spark-k8s-bundle/blob/c90143afafb701c409858b6ce91c2035010cc21e/.github/workflows/ci-tests.yaml#L95) some space and the tests works perfectly on the Github runners. We had some issues in the past week with self-hosted runners and Carl suggested to switch back to Github runners to avoid those instabilities in the tests.

PR #54, opened 2024-10-17
[DPE-5605] Add kyuubi HA in the bundle

welpaolo:
This PR add Kyuubi deployment with multiple units combined with load balancing done with Zookeeper.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -72,6 +72,7 @@ jobs:
           - snap_channel: "3.6/beta"
             agent: "3.6-beta2"
         exclude:
+          - backed: --backend yaml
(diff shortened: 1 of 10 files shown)


theoctober19th (approved):
LGTM

deusebio (approved):
LGTM!

I have one thoughts to be discussed in one comments, but to be addressed later, since `helpers.py` was not really part of this PR. 

Also I'm wondering if we should support both HA and non-HA in the bundle. I'm thinking about a configuration option true/false, that depending on its value we can either deploy and relate to ZooKeeper or not. Probably even just in the terraform, since it may be reasonably easy to be implemented. 

We can also have this into a separate ticket, more nice-to-have.

PR #68, opened 2024-12-15
[DPE-6242] Enabling running tests on existing deployments

deusebio:
This PR introduce the possibility to run tests against existing deployments. In order to make this possible a few tweaks were required:

* Introducing the [`skip_if_deployed`](https://github.com/canonical/spark-k8s-bundle/pull/68/files#diff-bcb3b0a9ca0a773899b801df1f8b1210cfbf23e90937a5a71f8a2b8fd592879bR39) marker for the parts about deployments
* Remove the cleanup of resources such as buckets, containers and service accounts
* Tune logs specific checks that would fail on long-running deployments, as [here](https://github.com/canonical/spark-k8s-bundle/pull/68/files#diff-d9f67f8d2266059cbc219044cb7859b89eccf934319a30da8229808f864e20a9R202)
* add CI for making sure that tests are idempotent and can be run against an existing deployments, see [here](https://github.com/canonical/spark-k8s-bundle/pull/68/files#diff-99c529469d54c618fa9ecca522a479a3e821e220c31d874dcf5a04fcfe876bd3)

PS. I have disabled all the tests for the sake of having faster (and cheaper CI) until we get to a close to ready-to-be merged state

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -44,7 +44,6 @@ jobs:
           cd python && tox -e unit
 
   integration-tests:
-    if: 'disabled' == 'true'
     name: Integration Tests
     runs-on: ubuntu-22.04
     timeout-minutes: 120
--- a/.github/workflows/ci-uat.yaml
+++ b/.github/workflows/ci-uat.yaml
@@ -105,7 +105,7 @@ jobs:
           tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
 
           tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
-          tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+          tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --model test-uat --cos-model cos
 
       # - name: Setup tmate session
       #   timeout-minutes: 20
@@ -210,7 +210,7 @@ jobs:
           tox run -e integration-sparkjob-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
 
           tox run -e integration-kyuubi-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
-          tox run -e integration-sparkjob-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+          tox run -e integration-sparkjob-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --model test-uat --cos-model cos
 
       # - name: Setup tmate session
       #   timeout-minutes: 20
--- a/.github/workflows/ci-uat.yaml
+++ b/.github/workflows/ci-uat.yaml
@@ -22,16 +22,16 @@ jobs:
           - snap_channel: "3.6/stable"
             agent: "3.6.0"
     steps:
-      # - name: (GitHub hosted) Free up disk space
-      #   timeout-minutes: 5
-      #   run: |
-      #     printf '\nDisk usage before cleanup\n'
-      #     df --human-readable
-      #     # Based on https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
-      #     rm -r /usr/share/dotnet
-      #     rm -r /opt/hostedtoolcache/
-      #     printf '\nDisk usage after cleanup\n'
-      #     df --human-readable
+      - name: (GitHub hosted) Free up disk space
+        timeout-minutes: 5
+        run: |
+          printf '\nDisk usage before cleanup\n'
+          df --human-readable
+          # Based on https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+          rm -r /usr/share/dotnet
+          rm -r /opt/hostedtoolcache/
+          printf '\nDisk usage after cleanup\n'
+          df --human-readable
       - name: Checkout repo
         uses: actions/checkout@v4
       - id: setup-python
@@ -88,9 +88,6 @@ jobs:
       - id: tests-integration
         name: Run Integration Tests
         timeout-minutes: 60
-        # env:
-        #   AZURE_STORAGE_ACCOUNT: ${{ secrets.AZURE_STORAGE_ACCOUNT }}
-        #   AZURE_STORAGE_KEY: ${{ secrets.AZURE_STORAGE_KEY }}
         run: |    
           cd python 
           
@@ -104,13 +101,9 @@ jobs:
           tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
           tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
 
+          # We re-run the last tests to make sure that the tests are idempotent
           tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
           tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --model test-uat --cos-model cos
-
-      # - name: Setup tmate session
-      #   timeout-minutes: 20
-      #   if: ${{ failure() }}
-      #   uses: mxschmitt/action-tmate@v3
 
   integration-tests-azure:
     name: Integration Tests (Azure)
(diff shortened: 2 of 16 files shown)


.github/workflows/ci-uat.yaml
```diff
@@ -0,0 +1,218 @@
...
+        #   AZURE_STORAGE_KEY: ${{ secrets.AZURE_STORAGE_KEY }}
+        run: |    
+          cd python 
+          
+          # The first tests deploy the bundle
+          tox run -e integration-bundle -- ${{ matrix.backend }} ${{ matrix.version }} --keep-models --model test-uat --cos-model cos
+
+          # Next we only use --no-deploy flag
+          
+          tox run -e integration-bundle -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
```
theoctober19th, line 102: I don't see the `no-deploy` flag used anywhere (and not in `conftest.py` where we parse these args..) Is this something automatically picked by ops_test?

theoctober19th: The same with the `--model` arg.

python/tests/integration/test_kyuubi.py
```diff
@@ -193,19 +193,31 @@ async def test_kyuubi_metrics_in_cos(ops_test: OpsTest, cos):
...
 
             # check for non empty logs
             assert len(logs) > 0
 
             # check if startup message is there...
-            assert any(
-                "Starting org.apache.kyuubi.server.KyuubiServer" in message
-                for timestamp, message in logs.items()
-            )
+            # assert any(
```
theoctober19th, line 202: Do we really want to remove these assertions?

python/tests/integration/test_kyuubi_azure.py
```diff
@@ -181,19 +181,30 @@ async def test_kyuubi_metrics_in_cos(ops_test: OpsTest, cos):
...
             assert len(logs) > 0
 
-            # check if startup message is there...
-            assert any(
-                "Starting org.apache.kyuubi.server.KyuubiServer" in message
-                for timestamp, message in logs.items()
-            )
+            # # check if startup message is there...
+            # assert any(
+            #     "Starting org.apache.kyuubi.server.KyuubiServer" in message
```
theoctober19th, line 191: Same comment as before, should we remove these assertions? (These are important to assert that the relevant logs are ingested into Loki in my opinion)

theoctober19th (commented):
I went through the PR and have a few things I want to understand regarding the implementation.

.github/workflows/ci-tests.yaml
```diff
@@ -44,6 +44,7 @@ jobs:
           cd python && tox -e unit
 
   integration-tests:
+    if: 'disabled' == 'true'
```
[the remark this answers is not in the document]
deusebio: We have many tests, I have disabled them for the time being, as per the comment in the PR

> PS. I have disabled all the tests for the sake of having faster (and cheaper CI) until we get to a close to ready-to-be merged state

.github/workflows/ci-uat.yaml
deusebio: yes, that's part of the [`pytest-operator`](https://github.com/charmed-kubernetes/pytest-operator/blob/main/docs/reference.md`)

python/tests/integration/test_kyuubi.py
welpaolo: Can we check this assertion conditionally to a new deployment only?

welpaolo (approved):
LGTM.  I agree with Bikalpa's comment. Can we conditionally check logs presence for new deployments ?

python/tests/integration/test_kyuubi.py
deusebio: The issue is that the query only fetch the latest logs, and for running deployments those do not include the KyuubiServer starting logs anymore. I would need to look into how to do the assertion only for new deployments, but maybe we could use other "more common" logs instead of the server starting, like the format of the logs generated during the query

deusebio: I'm thinking of re-enabling the log checks but switching the check from asserting the presence of `org.apache.kyuubi.server.KyuubiServer` to `org.apache.kyuubi.session.KyuubiSessionImpl"`

python/tests/integration/test_kyuubi_azure.py
deusebio: Addressed in [db4d894](https://github.com/canonical/spark-k8s-bundle/pull/68/commits/db4d894e7b462fec7b730b68e32e075fa6424178)

.github/workflows/ci-tests.yaml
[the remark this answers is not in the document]
deusebio: Re-enabled tests in [fa5fa4f](https://github.com/canonical/spark-k8s-bundle/pull/68/commits/fa5fa4f5f8ed4705bccfaf5b6c8ac6e2458b25c6)

.github/workflows/ci-uat.yaml
```diff
@@ -0,0 +1,221 @@
...
+          
+          tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+          tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+
+          tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
```
theoctober19th, line 107: This and the subsequent test is duplicated.

.github/workflows/ci-uat.yaml
```diff
@@ -0,0 +1,221 @@
...
+          cd python
+          
+          # The first tests deploy the bundle
+          tox run -e integration-bundle-azure -- ${{ matrix.backend }} ${{ matrix.version }} --keep-models --model test-uat --cos-model cos --uuid $UUID
+
+          # Next we only use --no-deploy flag
+          
+          tox run -e integration-bundle-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos --uuid $UUID
+          
+          tox run -e integration-kyuubi-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos --uuid $UUID
```
theoctober19th, line 212: This and the subsequent test is duplicated.

theoctober19th (approved):
LGTM! I see a few tests have been duplicated which I think can be cleaned up.

.github/workflows/ci-uat.yaml
deusebio: No no, I actually do this purposefully, to make sure that the tests are idempotent. One of the things we need for UAT is that they may be run multiple times. I had to fix a few things (like cleaning up the data), because running the tests twice would not work otherwise. I wanted to test this

.github/workflows/ci-uat.yaml
welpaolo: **nit**: Can you add a comment on for the test run two times ? Just to avoid this discussion again :)

.github/workflows/ci-uat.yaml
```diff
@@ -0,0 +1,221 @@
...
+          
+          tox run -e integration-bundle -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+          
+          tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+          tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+
+          tox run -e integration-kyuubi -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos
+          tox run -e integration-sparkjob -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --model test-uat --cos-model cos
+
+      # - name: Setup tmate session
```
welpaolo, line 110: Can you remove the commented lines ?

.github/workflows/ci-uat.yaml
```diff
@@ -0,0 +1,221 @@
...
+      matrix:
+        version:
+          - --spark-version "3.4.2"
+        backend:
+          - --backend yaml
+        juju:
+          - snap_channel: "3.6/stable"
+            agent: "3.6.0"
+    steps:
+      # - name: (GitHub hosted) Free up disk space
```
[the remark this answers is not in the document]
welpaolo: Is there any specific reason to non free up space disk?

.github/workflows/ci-uat.yaml
```diff
@@ -0,0 +1,221 @@
...
+          
+          tox run -e integration-bundle-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos --uuid $UUID
+          
+          tox run -e integration-kyuubi-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos --uuid $UUID
+          tox run -e integration-sparkjob-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos --uuid $UUID
+
+          tox run -e integration-kyuubi-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --keep-models --model test-uat --cos-model cos --uuid $UUID
+          tox run -e integration-sparkjob-azure -- ${{ matrix.backend }} ${{ matrix.version }} --no-deploy --model test-uat --cos-model cos --uuid $UUID
+
+      # - name: Setup tmate session
```
welpaolo, line 218: Same as above

.github/workflows/ci-uat.yaml
welpaolo: Same as above, can you add a comment for the repeated test just to avoid future confusion?

.github/workflows/ci-uat.yaml
deusebio: yes, that's sensible.

.github/workflows/ci-uat.yaml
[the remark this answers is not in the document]
deusebio: I guess it was not needed. But yes, why not. I will re-instantiate this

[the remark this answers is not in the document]
deusebio: Oh no, it does not work because of the upgrade 24.04 of canonical self-hosted runners. I will remove this

PR #70, opened 2025-01-10
[DPE-7003] Rename azure storage interface from 'azure' to 'azure-storage'

theoctober19th:


Batalex (changes requested):
We also use `azure-credentials` in the TF azure modules, would you mind changing it there as well? Thanks

theoctober19th:
@Batalex I've fixed the name in TF bundles as well. Can you please re-review?

Batalex (approved):

welpaolo (approved):
LGTM, I believe you also need to update the azure storage integrator charm revision to get the correct interface

theoctober19th:
@welpaolo You're right, I'll update them once we have those charms released to their latest revisions.

PR #78, opened 2025-01-31
[MISC] Upgrade tooling

Batalex:
## Changes

- Align the tooling with the rest of our projects -> ruff, poetry v2, dropping python 3.9
- Similar to what we had on `spark8t`, some of the rules were not enforced. Therefore, I had to add a bunch of not-so-great docstrings to make the linter happy after migrating to ruff
- ~~Changed the way we use our own fixtures in the tests. According to pytest, we should not import fixture, as this behaviour can be unreliable (and could break without notice since it's not supported)~~ Could not make it work as it should in the end


theoctober19th (approved):
LGTM. Thanks for the chore work, Alex!

deusebio (approved):
Really, many thanks Alex! This chore work is really a lot appreciated! 

I'm quite curious about the fixture issue, and that may indeed require a dedicated PR. The idea there would be to have some functionalities that we can easily re-use in tests both here in spark-k8s-bundle tests, but also in the other repositories. I'm wondering why this is a pattern to not be used. But in that case, we should possibly convert the fixture into some functions to be called, but in that case we may miss the context behaviour, with pre-logic and post-logic

Anyhow, let's discuss and try to find the way forward on this!

PR #80, opened 2025-02-11
[DPE-6563] Release spark-k8s-test python package to PyPI

theoctober19th:


.github/workflows/release-spark-test.yaml
```diff
@@ -0,0 +1,190 @@
...
+          architecture: x64
+      - name: Install poetry
+        run: |
+          pipx install poetry
+          poetry self add poetry-plugin-export
+      - id: install_environment
+        name: Set up build environment
+        run: |
+          poetry lock
+          poetry export -f requirements.txt --output requirements.txt
```
Batalex, line 53: **question**: Why do we need this step? I could not remember adding this to spark8t so I went and had a look, and yeah it's not there

Batalex (changes requested):
Overall LGTM, but I have a couple of requests if you don't mind:
- Would you mind renaming the `spark_tests` directory into `spark_k8s_tests` for consistency with the package name? (This will impact the `package` section in `pyproject.toml`
- I was on the verge of opening a quick PR to fix that, but maybe we might as well fix it here to avoid some reviewing overhead. In `tox.ini`, could you remove the `poetry run black` line in the `fmt` environment definition?

.github/workflows/release-spark-test.yaml
theoctober19th: This step comes from inside `make setup` step in `spark8t`.

Batalex: Oh, right, but yeah we don't need to export the txt file in any case

theoctober19th:
> I was on the verge of opening a quick PR to fix that, but maybe we might as well fix it here to avoid some reviewing overhead. In tox.ini, could you remove the poetry run black line in the fmt environment definition?

Sure, I'm happy to make that change ;)

> Would you mind renaming the spark_tests directory into spark_k8s_tests for consistency with the package name? (This will impact the package section in pyproject.toml)

I wouldn't mind at all! However I believe having the package module structure as it is with the new name wouldn't pose significant problem either, (specially when changing the name would require almost all test modules to be refactored to update their import statements). Do you still think this change is worth the effort?

.github/workflows/release-spark-test.yaml
theoctober19th: I'll remove that step altogether :)

Batalex:
> I wouldn't mind at all! However I believe having the package module structure as it is with the new name wouldn't pose significant problem either, (specially when changing the name would require almost all test modules to be refactored to update their import statements). Do you still think this change is worth the effort? I kinda feel `spark_tests` is short and handy compared to the longer `spark_k8s_tests` name as well. @Batalex

Yeah, you're right, let's forget about that

.github/workflows/release-spark-test.yaml
theoctober19th: Removed in [e9754c2](https://github.com/canonical/spark-k8s-bundle/pull/80/commits/e9754c2107984cf5f8408fadf6d0a6493f43d3e3)

Batalex (approved):

welpaolo (approved):
LGTM

PR #81, opened 2025-02-11
[DO NOT MERGE] Debug pending service in CI

Batalex:


Batalex:
Closing this, as we found out the reason for this issue: metallb was only using 1 IP for external access.

PR #84, opened 2025-02-12
[DPE-6570] Stabilize CI

Batalex:
## Changes

- Upgrade microk8s to 1.32/stable
- Upgrade minimum juju version to 3.5
- Assign a bigger IP pool to metallb for K8s services
- Bump kyuubi to rev 39 (prevent container access while unreachable)
- Cosmetics: improved the job name so that we can see the test matrix in GH interface
- Increase # of parallel tests from 7 to 15

### DIFF
--- a/.github/workflows/release-spark-test.yaml
+++ b/.github/workflows/release-spark-test.yaml
@@ -18,6 +18,7 @@ jobs:
   tests:
     name: Run CI tests
     uses: ./.github/workflows/ci-tests.yaml
+    secrets: inherit
 
   release_checks:
     name: Checks before release


theoctober19th (approved):
The changes look good and sane to me. Only thing I'm rather not sure about would be the consequences of increasing the number of parallel runners. Only yesterday I faced a strange scenario where I had to wait for almost 2 hours for a runner to be assigned. I'm not much knowledgeable on how the runner allocation works in GitHub internally, however my point is that let's keep on eye on that as well.

Batalex:
> The changes look good and sane to me. Only thing I'm rather not sure about would be the consequences of increasing the number of parallel runners. Only yesterday I faced a strange scenario where I had to wait for almost 2 hours for a runner to be assigned. I'm not much knowledgeable on how the runner allocation works in GitHub internally, however my point is that let's keep on eye on that as well.

Yeah I saw that, but it was a self-hosted runner, not a GH one, right?

PR #87, opened 2025-02-24
[MISC] Release to TestPyPI on pull request

theoctober19th:


.github/workflows/publish-to-pypi.yaml
```diff
@@ -96,9 +92,10 @@ jobs:
           name: distfiles
           path: python/dist/
 
+
   autorelease:
     name: Release the package on github
-    needs: [release_checks, tests, build]
```
Batalex, line 101: **question**: Do we need the `tests` job if we remove it as a constraint everywhere?

.github/workflows/publish-to-testpypi.yaml
```diff
@@ -0,0 +1,109 @@
...
+          poetry build
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v4
+        with:
+          name: distfiles
+          path: python/dist/
+
+  upload-to-test-pypi:
+    name: Publish to PyPI
+    needs: [release_checks, build]
```
Batalex, line 92: **todo**: Even if it is aimed at testpypi, we should make sure that the unit tests run fine before uploading anything.

Actually, we also need it in the pypi publishing job as well

python/spark_test/core/kyuubi.py
```diff
@@ -3,15 +3,15 @@
...
+from typing import Iterable, Union
 
 from pyhive.hive import Connection
 
 TYPES = {int: "int", str: "string"}
 
 ITYPES = {v: k for k, v in TYPES.items()}
 
-SchemaType = Type[int | str]
+SchemaType = Union[int | str]
```
Batalex, line 14: **suggestion**: let's use the 3.10 syntax

```py
from typing import TypeAlias

SchemaType: TypeAlias = int | str
```

python/spark_test/core/kyuubi.py
```diff
@@ -29,23 +29,28 @@ class TableNotFound(Exception):
...
 
-    def __init__(self, username: str, password: str, host: str, port: int = 10009):
-        self.username = username
-        self.password = password
+    def __init__(
+        self,
+        host: str = "localhost",
+        port: int = 10009,
+        username: str = None,
+        password: str = None,
```
Batalex, line 37: **todo**: the types should be `str | None`

python/spark_test/utils.py
```diff
@@ -3,14 +3,20 @@
...
 from lightkube import Client
 from lightkube.resources.core_v1 import Pod as LightKubePod
 
 from spark_test.core.pod import Pod
 
+SPARK_EXECUTOR_POD_REGEX = r".*-exec-\d+"
+
 
-def get_spark_drivers(client: Client, namespace: str) -> list[Pod]:
+def get_spark_drivers(client: Client = None, namespace: str = "default") -> list[Pod]:
```
Batalex, line 16: **todo**: The type is wrong: `Client | None` 

Also in `get_spark_executors`

.github/workflows/publish-to-pypi.yaml
welpaolo: Maybe have `tests` will be a good idea if the new features are executed were tests job is executed

python/spark_test/utils.py
theoctober19th: Fixed in [766e8a6](https://github.com/canonical/spark-k8s-bundle/pull/87/commits/766e8a668a5e79fe1fcdb57ab09d2a1be60ed525)

python/spark_test/core/kyuubi.py
theoctober19th: Fixed in [766e8a6](https://github.com/canonical/spark-k8s-bundle/pull/87/commits/766e8a668a5e79fe1fcdb57ab09d2a1be60ed525).

python/spark_test/core/kyuubi.py
theoctober19th: Fixed in [766e8a6](https://github.com/canonical/spark-k8s-bundle/pull/87/commits/766e8a668a5e79fe1fcdb57ab09d2a1be60ed525).

.github/workflows/publish-to-testpypi.yaml
theoctober19th: The PyPI publishing job runs it through `ci-test-minimal` which in turn calls `ci-checks` that does the unit tests. For the TestPyPI, I have just added `ci-checks` step before publishing. Thanks -- fixed in [766e8a6](https://github.com/canonical/spark-k8s-bundle/pull/87/commits/766e8a668a5e79fe1fcdb57ab09d2a1be60ed525).

.github/workflows/publish-to-pypi.yaml
theoctober19th: No, we need not remove it. This was removed accidentally when I moved code to this workflow from the other one. Fixed in [766e8a6](https://github.com/canonical/spark-k8s-bundle/pull/87/commits/766e8a668a5e79fe1fcdb57ab09d2a1be60ed525).

Batalex (approved):

PR #88, opened 2025-02-25
[DPE-6620] Initial benchmark implementation

Batalex:
This PR is the first attempt at implementing a way to stress test our solution in a reproducible way.

We use the [TPC-H connector](https://kyuubi.readthedocs.io/en/v1.9.3/connector/spark/tpch.html#tpc-h-integration) for Kyuubi to generate synthetic data without access to an external data source (or cluttering the repo with test data).

## Changes

- Added a tox environment and a test file to run benchmarks against the TPC-H dataset
- Added a pytest option to control the data size factor
- Create a simple html report 

What's next?
- Our current setup in this PR do not use TLS nor external access. The benchmarks are not I/O intensive though, so I don't think that they are not representative even with cipher + k8s service adding latency
- If we want to run the queries multiple times, we can avoid cache issues by mutating a little bit the queries (e.g. changing slightly a parameter = we assume no impact on performance, but it disables the cache for a fair comparison over multiple loop iteration). This would require some more tooling though.
- Come up with an experiment plan (fixed hardware) so that we can have our baseline.

This is what the report looks like (created here on a GH runner):
image

This is quite difficult to assess right now if we have a cache issue requiring mutating the queries. If we had, we would expect each first run to be significantly longer than the rest, but this is not the case here.

### DIFF
--- a/python/tests/integration/test_benchmark_kyuubi.py
+++ b/python/tests/integration/test_benchmark_kyuubi.py
@@ -36,6 +36,12 @@ def bench_iterations(request) -> int:
     return request.config.getoption("--bench-iterations")
 
 
+@pytest.fixture(scope="module")
+def report_name(request) -> str:
+    """Get benchmark number of iterations."""
+    return request.config.getoption("--report-name")
+
+
 @pytest.mark.skip_if_deployed
 @pytest.mark.abort_on_fail
 async def test_deploy_bundle(ops_test: OpsTest, spark_bundle) -> None:
@@ -116,7 +122,7 @@ async def test_setup_env(ops_test: OpsTest, sf: str) -> None:
 
 @pytest.mark.abort_on_fail
 async def test_run_benchmark_queries(
-    ops_test: OpsTest, sf: str, bench_iterations: int
+    ops_test: OpsTest, sf: str, bench_iterations: int, report_name: str
 ) -> None:
     """Run benchmark queries and generate report."""
     logger.info("Running benchmark queries")
@@ -170,7 +176,7 @@ async def test_run_benchmark_queries(
         .tab_source_note(md("*a `-1` value in the spark line denotes a missing value."))
     )
 
-    table.write_raw_html("report.html")
+    table.write_raw_html(f"{report_name}.html")
     logger.info("Report written to 'report.html'")
 
 
--- a/python/tests/integration/test_benchmark_kyuubi.py
+++ b/python/tests/integration/test_benchmark_kyuubi.py
@@ -114,4 +114,4 @@ async def test_run_benchmark_queries(ops_test: OpsTest):
             cursor.execute(stmt)
             cursor.fetchall()
             end = time.monotonic()
-            logger.info(f"TPC-H #1: {end - start}")
+            logger.info(f"TPC-H #{i}: {end - start}")
--- a/python/tests/integration/test_benchmark_kyuubi.py
+++ b/python/tests/integration/test_benchmark_kyuubi.py
@@ -135,10 +135,10 @@ async def test_run_benchmark_queries(
                     cursor.execute(stmt)
                     cursor.fetchall()
                 except Exception:
-                    report_data[i].append(None)
+                    report_data[str(i)].append(None)
                 else:
                     end = time.monotonic()
-                    report_data[i].append(end - start)
+                    report_data[str(i)].append(end - start)
 
     logger.info("Creating report")
     df = pl.DataFrame(report_data)
--- a/python/tests/integration/test_benchmark_kyuubi.py
+++ b/python/tests/integration/test_benchmark_kyuubi.py
@@ -67,13 +67,6 @@ async def test_setup_env(ops_test: OpsTest, sf: str) -> None:
     assert ops_test.model is not None
 
     logger.info("Setup TPC-H connector")
-    check_output(
-        f"JUJU_MODEL={ops_test.model_full_name} juju remove-relation {KYUUBI} {HUB}",
-        stderr=PIPE,
-        shell=True,
-        universal_newlines=True,
-    )
-    await ops_test.model.wait_for_idle(apps=[KYUUBI, HUB], idle_period=20, timeout=600)
 
     leader_unit_id = await get_leader_unit_number(ops_test, HUB)
     logger.info(f"Leader unit: {HUB}/{leader_unit_id}")
@@ -88,7 +81,6 @@ async def test_setup_env(ops_test: OpsTest, sf: str) -> None:
     )
     await action.wait()
 
-    await ops_test.model.add_relation(KYUUBI, HUB)
     async with ops_test.fast_forward(fast_interval="90s"):
         await ops_test.model.wait_for_idle(
             apps=[KYUUBI, HUB],
@@ -205,12 +197,4 @@ async def cleanup(ops_test: OpsTest) -> None:
         conf="spark.jars.packages",
     )
     await action.wait()
-    check_output(
-        f"JUJU_MODEL={ops_test.model_full_name} juju remove-relation {KYUUBI} {HUB}",
-        stderr=PIPE,
-        shell=True,
-        universal_newlines=True,
-    )
-    await ops_test.model.wait_for_idle(apps=[KYUUBI, HUB], idle_period=20, timeout=600)
-    await ops_test.model.add_relation(KYUUBI, HUB)
     await ops_test.model.wait_for_idle(apps=[KYUUBI, HUB], idle_period=20, timeout=600)
--- a/python/tests/integration/test_benchmark_kyuubi.py
+++ b/python/tests/integration/test_benchmark_kyuubi.py
@@ -2,10 +2,14 @@
 # Copyright 2025 Canonical Ltd.
 # See LICENSE file for licensing details.
 import asyncio
+from collections import defaultdict
 import logging
 import time
 from subprocess import PIPE, check_output
 
+import polars as pl
+from great_tables import GT, md, loc, style
+from datetime import date
 import pytest
 from pyhive.hive import Cursor
 from pytest_operator.plugin import OpsTest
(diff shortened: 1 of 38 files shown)


Batalex:
I'll keep the PR as draft as long as we don't have the tpc-h connector in the image. Mostly looking for feedback and validation before we commit to adding it

Batalex:
> I've been struggling to understand how the job is triggered in CI. How do we do this? Shouldn't we have a workflow that is callable? I would not put this in regular CI (like pull-request), probably not even scheduled

So, while I was working on this, I hijacked the `on_pull_request` workflow to ignore the test and run the new tox env, but I reverted those changes when I was done.

I don't think that this benchmark should run on GH runners because the hardware is not fixed from one run to the next. I am not even sure that within one run we don't get some performance fluctuation because of the load on the actual physical machine. That would certainly explain how, on query # 7 for instance, we see the second run taking twice as long as the first.
IMO we should clone the repo in the deployment to be tested, and then run the benchmark tool, similar to what we do with user acceptance tests

python/tests/integration/test_benchmark_kyuubi.py
```diff
@@ -0,0 +1,207 @@
...
+    leader_unit_id = await get_leader_unit_number(ops_test, HUB)
+    logger.info(f"Leader unit: {HUB}/{leader_unit_id}")
+    action = await ops_test.model.units.get(f"{HUB}/{leader_unit_id}").run_action(
+        "add-config",
+        conf="spark.sql.catalog.tpch=org.apache.kyuubi.spark.connector.tpch.TPCHCatalog",
+    )
+    await action.wait()
+    action = await ops_test.model.units.get(f"{HUB}/{leader_unit_id}").run_action(
+        "add-config",
+        conf="spark.kubernetes.container.image=docker.io/batalex/charmed-spark-kyuubi:1",
```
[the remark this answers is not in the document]
Batalex: The connector jar is like 3MB, so it represents less than a 1% size increase to the image. On the other hand, we probably want to keep the image as minimal as possible, both in size and attack surface. I did not check if the connector had any CVE, but we most likely want to keep our good track record :D

Let's have a new set of "-test" images

deusebio:
Yes, that's fair. Maybe we could just run these against a EKS or Azure deployments instead. But yes, the consideration is sensible and I'm not pushing for having this on CI. But in this case we should probably have a section in the readme or a dedicated docs for this such that we can use it easily. I kinda also use CI as documentation, hence why I was asking to see its usage

theoctober19th:
Solid work here, well done @Batalex! I quickly went through the PR and this time I'm leaving out code structuring and implementation details (since I realize from the conversation that we are kinda in the draft stage).

One thing I wanted to see in addition is what is done in [the docker image] you mentioned above in PR description. Is it just adding the JARs? 

If so, we may add those packages on-the-fly using property like `spark.jars.packages` during the test -- which will pull those packages from Maven and stuff those inside driver and executor classpaths. Doing so, we may not have the need to release a "test" spark image.

python/tests/integration/test_benchmark_kyuubi.py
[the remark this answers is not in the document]
theoctober19th: One thing I wanted to see in addition is what is done in [the docker image] you mentioned above in PR description. Is it just adding the JARs? 

If so, we may add those packages on-the-fly using property like `spark.jars.packages` during the test -- which will pull those packages from Maven and stuff those inside driver and executor classpaths. Doing so, we may not have the need to release a "test" spark image.

theoctober19th (commented):
Solid work here, well done @Batalex! I quickly went through the PR and this time I'm leaving out code structuring and implementation details (since I realize from the conversation that we are kinda in the draft stage).

One thing I wanted to see in addition is what is done in [the docker image] you mentioned above in PR description. Is it just adding the JARs?

If so, we may add those packages on-the-fly using property like spark.jars.packages during the test -- which will pull those packages from Maven and stuff those inside driver and executor classpaths. Doing so, we may not have the need to release a "test" spark image.

Batalex:
So, @theoctober19th's suggestion worked flawlessly, so here is the update PR ready for review. We do not depend on a custom image anymore, so we are good to go with this initial implementation! 🎉

python/tests/integration/test_benchmark_kyuubi.py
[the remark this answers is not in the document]
Batalex: `spark.jars.packages` worked perfectly. Changed in 01284d0b03365ab0e57f5bb46f1f77240f76a9a9

python/tests/integration/test_benchmark_kyuubi.py
```diff
@@ -0,0 +1,210 @@
...
+import pytest
+from great_tables import GT, loc, md, style
+from pyhive.hive import Cursor
+from pytest_operator.plugin import OpsTest
+
+from spark_test.core.kyuubi import KyuubiClient
+
+from .helpers import get_kyuubi_credentials, get_leader_unit_number
+
+KYUUBI = "kyuubi"
```
welpaolo, line 21: Should we make the name of the two applications (kyuubi and integration-hub) configurable from a fixture ?

python/tests/integration/test_benchmark_kyuubi.py
```diff
@@ -0,0 +1,210 @@
...
+        cursor.execute("DROP DATABASE bench;")
+
+    logger.info("Cleaning bench config")
+    leader_unit_id = await get_leader_unit_number(ops_test, HUB)
+    action = await ops_test.model.units.get(f"{HUB}/{leader_unit_id}").run_action(
+        "remove-config",
+        key="spark.sql.catalog.tpch",
+    )
+    await action.wait()
+    action = await ops_test.model.units.get(f"{HUB}/{leader_unit_id}").run_action(
```
welpaolo, line 197: Why do you need to remove the image ?

welpaolo (approved):
Looks good to me! Just left some minor comments

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Oops, this is a remnant of an older setup. I originally made a custom kyuubi image, but then Bikalpa pointed that I could just use `spark.jars.packages` to add the connector jar

Batalex: Actually, we should do something else here. Since the ultimate goal is to have this as an "extension" of the user acceptance tests, we should configure kyuubi to use a different service account, add the benchmark config there, run the bench, and then clean up by re-configuring the old service account and finally deleting the test one

python/tests/integration/test_benchmark_kyuubi.py
```diff
@@ -0,0 +1,210 @@
...
+        .fmt_nanoplot(columns="raw")
+        .fmt_number(columns=pivoted.columns[1:], decimals=2, drop_trailing_zeros=True)
+        .cols_label(query="Query #", raw="Iteration time*")
+        .tab_style(
+            style=style.text(transform="capitalize"), locations=loc.column_header()
+        )
+        .tab_source_note(md("*a `-1` value in the spark line denotes a missing value."))
+    )
+
+    table.write_raw_html("report.html")
```
welpaolo, line 173: Can we also have the report name configurable ?

python/tests/integration/test_benchmark_kyuubi.py
deusebio: +1 on having a orthogonal service account to make it very easy to cleanup

deusebio (approved):
I don't have big concerns on the PR. If it does not change the scope of the PR significantly, I would address your own sugggestion [here](https://github.com/canonical/spark-k8s-bundle/pull/88/files#r2035776067)

python/tests/integration/test_benchmark_kyuubi.py
theoctober19th: +1 on a separate service account. 

However I'm not sure we currently have support for hot swapping service account in the Kyuubi charm. Currently you would need to change the config to change the service account for Kyuubi, however this does not trigger relation changed for the Integration Hub to create this new service account and supply to Kyuubi. Moreover, even if we add this trigger, Integration Hub will complain the service account already exists when we hot swap the service account to the previous one after the benchmark tests are done. This is somehow also related to what Alex commented [in this PR](https://github.com/canonical/data-platform-libs/pull/208#pullrequestreview-2750254287).

python/tests/integration/test_benchmark_kyuubi.py
```diff
@@ -0,0 +1,210 @@
...
+        conf="spark.sql.catalog.tpch=org.apache.kyuubi.spark.connector.tpch.TPCHCatalog",
+    )
+    await action.wait()
+    action = await ops_test.model.units.get(f"{HUB}/{leader_unit_id}").run_action(
+        "add-config",
+        conf="spark.jars.packages=org.apache.kyuubi:kyuubi-spark-connector-tpch_2.12:1.9.3",
+    )
+    await action.wait()
+
+    await ops_test.model.add_relation(KYUUBI, HUB)
```
theoctober19th, line 85: Wouldn't this relation be already there when the bundle is deployed?

theoctober19th (commented):
Solid work here @Batalex! 

I have a few concerns that I have left in the comments.

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Here is how it would play out:

1. The bundle is deployed, possibly with a custom setup (for instance, in a customer deployment). We already have a relation between Kyuubi and the Hub named after the current value for the charm configuration option.
2. We backup the spark properties (and the service account, let's say `default:kyuubi-engine`) in a file
3. We remove the relation, the service account `default:kyuubi-engine` is removed by the hub
4. We change the configuration option to something else, like `uat:sa-bench`
5. We relate Kyuubi and the Hub, `uat:sa-bench` is created by the hub
6. We restore the previous configuration from `default:kyuubi-engine` and add the tpch catalog as well as the jar.
7. We run the bench
8. We remove the relation once again, `uat:sa-bench` is removed
9. We change the config option back to `default:kyuubi-engine`
10. We relate for the final time, `default:kyuubi-engine`  is created
11. We restore the properties from the file
12. We cleanup the files

As you can see, this process is kinda hacky, but I believe it allows us to clean up nicely the deployment. Since it's quite different from what we have here, let me know with a :+1: if I should proceed.
We could simplify this a little bit if we allow a charm to request an existing service account, though

theoctober19th: This is totally out of the scope of this PR. Even if we do this (which I don't think likely), let's not do this in this PR.

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Could you tell me more about how we could do that? Did you mean adding a new parser option, or looking through the deployed applications for a match?

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Done in 82bede656a3ece99fd7e77829b4a262ae754c917

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Cleaned up `spark.jar.packages` in 9acc5b2a7a751e06820742212bbc6beb9e72065d. We can figure out the rest later for now

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Yes, since we agreed on going with the most straightforward way for now, bumping is better

Batalex: Done in cf75e5989e375a2820d1e4289ce8bcbc1e4995cb

python/tests/integration/test_benchmark_kyuubi.py
Batalex: Done in 3cb12e7

PR #92, opened 2025-03-06
[DPE-6425] Disable HA Zookeeper for tests

Batalex:
## Changes

- Did a little bit of cleanup on the TF bundles: removed leftover file from previous structure and renamed 'modules' directory, similar to what we have elsewhere in the company
- Removed tmate step in ci, following prior discussion
- Parametrized the number of ZK units (still defaults to 3). We certainly don't need HA for integration test, and we will appreciate the test speed up and freeing some resources on the machine

According to what I saw, this change does reduce the number of failures related to ZK, though I still experienced one with a single unit failing to elect a leader. The tests are still flaky, though, we get plenty of errors because of a relation-changed failed hook on the logging endpoint. Not sure what we can do on our end, since everything is handled by the loki lib


deusebio (approved):
Looks good! These are all sensible modifications, and I believe it is sane to expose the number of zk units as a variable. Actually we should probably provide a bit more customization for the modules, e.g. we should allow users to specify channels or revisions possibily, as this may be a constraint coming as output of the spec on TF being worked on by the working group. @marcoppenheimer should be our main point of contact there

> The tests are still flaky, though, we get plenty of errors because of a relation-changed failed hook on the logging endpoint. Not sure what we can do on our end, since everything is handled by the loki lib

Have we raised this with the COS team? If they are as puzzled as we are and they don't have a recommendation how to fix this flakyness, we can raise an issue on their repo and link it in the comment/code here

PR #79, opened 2025-02-03
[DPE-6299] Port updated terraform bundle

Batalex:
This PR ports the work done on the managed solutions' repository.

## Changes:
- New modular structure for terraform bundle, both for spark 3.4 and 3.5
- Updated revisions (charms that are part of the bundle, OCI image). See table below
- Enable terraform deployment when using azure storage

## To be done / Issues:
- Figure out why the TLS relation prevents Kyuubi from starting
- Replace `self-signed-certificates` (with the manual provider charm?)
- Good to know, I tried to upgrade PostgreSQL charms during my initial copy/paste and got an error because new releases are incompatible with juju 3.4.2.

We have some flakiness on the CI, which becomes a problem considering the massive number of tests that we run:

-   (rare) S3 credentials incorrect
-   **FIXED**: Kyuubi waiting for pebble after a faulty restart (happens during `_on_zookeeper_changed`, similar to what I saw with TLS). We might need to check guard some operations, like deferring if the container is not ready for a file write operation. Here is a truncated trace:

<details><summary> Trace </summary>

```
2025-02-07T14:41:29.584Z [container-agent]   File "/var/lib/juju/agents/unit-kyuubi-0/charm/src/events/zookeeper.py", line 43, in _on_zookeeper_changed
2025-02-07T14:41:29.584Z [container-agent]     self.kyuubi.update()
2025-02-07T14:41:29.584Z [container-agent]   File "/var/lib/juju/agents/unit-kyuubi-0/charm/src/managers/kyuubi.py", line 62, in update
2025-02-07T14:41:29.584Z [container-agent]     self._compare_and_update_file(
2025-02-07T14:41:29.584Z [container-agent]   File "/var/lib/juju/agents/unit-kyuubi-0/charm/src/managers/kyuubi.py", line 31, in _compare_and_update_file
2025-02-07T14:41:29.584Z [container-agent]     existing_content = self.workload.read(file_path)
2025-02-07T14:41:29.584Z [container-agent]   File "/var/lib/juju/agents/unit-kyuubi-0/charm/src/common/workload/k8s.py", line 45, in read
2025-02-07T14:41:29.584Z [container-agent]     if not self.container.exists(path):
2025-02-07T14:41:29.584Z [container-agent]   File "/var/lib/juju/agents/unit-kyuubi-0/charm/venv/lib/python3.10/site-packages/ops/model.py", line 2852, in exists
2025-02-07T14:41:29.584Z [container-agent]     self._pebble.list_files(str(path), itself=True)
[...]

PR #94, opened 2025-03-11
[DPE-6712] Extract external TF modules

Batalex:
To be merged after #93 

Following up on a comment from Enrico in the initial TF bundle upgrade PR #79. 
We want to logically separate the parts of the bundle where we have full control/ownership over parts maintained by others. This is highlighted in the new TF bundle READMEs with the following minimal example:

```hcl
module "ssc" {
  source      = "git::https://github.com/canonical/self-signed-certificates-operator//terraform"
  model       = juju_model.spark.name
  channel     = "latest/edge"
  revision    = 163
  constraints = "arch=amd64"
  base        = "ubuntu@22.04"
  units       = 1
}

module "spark" {
  depends_on                = [juju_model.spark]
  source                    = "./releases/3.5/terraform/modules/spark"
  model                     = juju_model.spark.name
  tls_app_name              = module.ssc.app_name
  tls_certificates_endpoint = module.ssc.provides.certificates
}

module "azure" ...
```

Using this structure, a customer deployment can choose the manual TLS operator and still use the bundle as-is

## Changes

- Extracted COS and TLS out of the "core" spark bundle since we are not in charge of maintaining those. Relations / offers are now explicitly passed as module variables, with proper documentation.
- Updated TF bundle documentation


deusebio:
This is really good piece of work @Batalex ! I haven't reviewed the code yet, as the PR is rather big because it includes also the changes in #93 , however by browsing the code I have a couple of questions:

1. For cos, we bind to offers, whereas for tls we bind to app/endpoints. Why not using offers for tls too? I'm wondering if we want to also support the use case where the tls operator is deployed in a different model, and then exposed through offers. Would there be a way to support both? Do offer allowed to be exposed in the same model?
2. Shouldn't we move the external modules in the tests too (or even just in the tests, and in the tests have the uber TF module that can be deployed?)

Batalex:
> Why not using offers for tls too?

That's an interesting thought, for sure. We have always deployed `self-signed-certificates` alongside the charms that would use certs, but that was absolutely a convenience rather than a requirement.
I think we can play around with terraform meta arguments and assertions to have either one of `tls_relation` / `tls_offer` and properly wire everything. I'll give it a try.

> Shouldn't we move the external modules in the tests?
This makes sense, I'll take care of that

Batalex:
I have been unsuccessful so far in moving the `cos` and `tls` modules to the tests resources because this requires some pretty hefty changes to the way we load the bundle in the tests fixtures (where we copy the content of the release folder into a tmp folder before running terraform commands).

deusebio:
> I have been unsuccessful so far in moving the cos and tls modules to the tests

I honestly believe the current separation is already bringing a good value to separate the concerns, and I can see why refactoring the tests may require some further work we can just place into a dedicated ticket. I'm happy with the current status

deusebio (approved):

theoctober19th (approved):
Great work, Alex! Let's merge this.

PR #96, opened 2025-03-25
[DPE-6889] Address issues found during the integration sprint

Batalex:
## Changes

- Fixed an issue where we were forgetting the wrong model alias, or using the alias instead of the actual name in a subprocess
- Fixed test about finding job logs in object storage by getting the information from the integration hub. The original version was fine in the context of this repository's CI, but when we run the integration tests as user acceptance tests, a different object storage from the one created by the test is configured on the hub. 
- Fixed CI failing by pinning scrape-config to rev 51. We should open a new ticket to figure out what is going on with this revision. The issue could be that the metrics are scrapped every minute because the 10s interval configuration option is not respected

### DIFF
--- a/python/tests/integration/conftest.py
+++ b/python/tests/integration/conftest.py
@@ -343,7 +343,7 @@ async def cos(ops_test: OpsTest, cos_model: str, backend: str):
 
         yield cos_model
 
-        await ops_test.forget_model(cos_model)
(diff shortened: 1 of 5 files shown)


deusebio (approved):
LGTM!!!

theoctober19th (approved):

PR #102, opened 2025-04-24
[MISC] Add terraform resources to bootstrap AKS cluster for Charmed Spark

theoctober19th:
This PR adds the TF module to bootstrap AKS cluster from [here](https://github.com/theoctober19th/charmed-spark-aks#) to this repository.


releases/3.4/yaml/docs-resources/aks-setup/.gitignore
Batalex, line 1: **todo**: Let's merge the rules (if we want to keep them) with the top level `.gitignore`

releases/3.4/yaml/docs-resources/aks-setup/LICENSE
Batalex, line 1: **question**: I don't have the answer TBH, but I don't think this licence should be left here as-is. Especially when the rest of the codebase is under Apache and every file has a corporate copyright header

theoctober19th: I agree, I'll remove the LICENSE file -- this got copied when I copied the code from the other repo.

releases/3.4/yaml/docs-resources/aks-setup/.gitignore
theoctober19th: I kinda don't like the idea of merging these two `.gitignore`s specially because I don't want this resource Terraform bundle to have any more footprint than this folder itself. Why do you think merging them would be a good idea?

welpaolo (approved):
LGTM, please adjust the license as suggested by @Batalex

releases/3.4/yaml/docs-resources/aks-setup/.gitignore
Batalex: Let's put it a different way: it's not like we want to track any `tfstate` file in the parent folders either, but any time we are working on other TF modules we are at risk of committing such a file

izmalk (commented):
I see no problem in this PR, but I lack experience using terraform to explicitly approve it. LGTM. Let me know if you still need the official approval.

Batalex (approved):

PR #108, opened 2025-05-15
[DPE-7332] Jubilant migration

Batalex:
## Changes

- No longer any async code in the tests :tada: 
- The cascading fixtures to deploy the bundle should be a little simpler to follow.


python/tests/integration/bench/test_benchmark_kyuubi.py
```diff
@@ -1,73 +1,103 @@
...
+KYUUBI = "kyuubi"
+HUB = "integration-hub"
 
 logger = logging.getLogger(__name__)
 
 
+def pytest_addoption(parser):
+    """Add CLI options to pytest."""
+    parser.addoption(
+        "--foo",
```
theoctober19th, line 28: nit: can we pick a more specific name than `foo`? Something like `benchmark_report_filename`  for example?

python/tests/integration/helpers.py
```diff
@@ -442,31 +431,9 @@ async def deploy_bundle_terraform(
     return list(outputs["charms"]["value"].values())
 
 
-async def add_juju_secret(
```
theoctober19th, line 445: question: Isn't this still relevant for deploying YAML bundles? (since I think we have no way to create secrets in when deploying YAML and thus need to do it manually?)

python/tests/integration/test_benchmark_kyuubi.py
```diff
@@ -1,218 +0,0 @@
-#!/usr/bin/env python3
```
theoctober19th, line 1: question: why do we delete this entire test file?

theoctober19th (commented):
I have a few questions -- otherwise the overall PR looks good.

python/tests/integration/test_benchmark_kyuubi.py
Batalex: It's a duplicate from `python/tests/integration/test_benchmark_kyuubi.py` that we missed during review

python/tests/integration/bench/test_benchmark_kyuubi.py
Batalex: oops, development leftover. We do have a `report-name` option/fixture
Remove in 5e00a0b3fc083bdba375710286c8cd2e9fcbf4fa

python/tests/integration/helpers.py
Batalex: The operation is still done manually in `integration/conftest.py L501 for both yaml and tf, it's just that jubilant make is so easy that we don't need a dedicated helper that adds unneeded overhead

theoctober19th (approved):

PR #109, opened 2025-05-20
[DPE-7333] Unpin boto3

Batalex:
We need to release a new spark-k8s-test without the upper constraint on boto3 to make https://github.com/canonical/kyuubi-k8s-operator/pull/81 possible

## Changes
- Add boto config to restore compatibility with third party providers; thus removing the pin in the process
- Adjust the logging and timeout config of jubilant for a better debugging experience on CI

### DIFF
--- a/python/tests/integration/conftest.py
+++ b/python/tests/integration/conftest.py
@@ -50,7 +50,7 @@ COS_APPS = [
 ]
 FORWARD_TIMEOUT_SECONDS = 10
 logger = logging.getLogger(__name__)
-logging.getLogger("jubilant.wait").setLevel(logging.WARNING)
+# logging.getLogger("jubilant.wait").setLevel(logging.WARNING)
 
 
 def pytest_runtest_setup(item):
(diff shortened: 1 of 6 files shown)


python/tests/integration/conftest.py
```diff
@@ -50,7 +50,7 @@
 ]
 FORWARD_TIMEOUT_SECONDS = 10
 logger = logging.getLogger(__name__)
-logging.getLogger("jubilant.wait").setLevel(logging.WARNING)
+# logging.getLogger("jubilant.wait").setLevel(logging.WARNING)
```
welpaolo, line 53: Why this line is commented?

welpaolo (approved):
LGTM, I just left a minor comment

python/tests/integration/conftest.py
Batalex: Actually, I'd like to have your input on that.
Would you mind having a look at the CI logs and share your thoughts on the verbosity?

With this line enabled, we get 0 logs until the `juju.wait` times out (30 minutes) if anything goes wrong with the deployment.
Commented, we can see if a deployment is stuck (and where), which makes it easier to debug. But the logs are 1000x bigger

theoctober19th (approved):
LGTM

python/tests/integration/conftest.py
welpaolo: I will prefer to have more logs to understand quickly the issue. I don't see the amount of logs as a big issue for the moment.

PR #111, opened 2025-05-22
[DPE-7336] Enable COS deployment and charm revisions specification

deusebio:
The PR provides a way to customize the COS deployment, as described in [DPE-7336](https://warthogs.atlassian.net/browse/DPE-7336), as well as enabling charm revision specifications, following [DA183](https://docs.google.com/document/d/11mAVQgEbTlyn3YwM9--4DxrNec_3BQ-zoYqlgFQ4RLY/edit?tab=t.0)

[DPE-7336]: https://warthogs.atlassian.net/browse/DPE-7336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


releases/3.4/terraform/variables.tf
```diff
@@ -30,29 +30,25 @@ variable "create_model" {
...
-  type        = string
-  default     = ""
-  description = "COS certificate"
-}
-variable "COS_TLS_KEY" {
-  type        = string
-  default     = ""
-  description = "COS certificate key"
-}
-variable "COS_TLS_CA" {
```
deusebio, line 52: **question** I'm not sure where these variables were used actually. I have addede them to the "cos" object, but I'm not sure what we should do with them

releases/3.4/terraform/variables.tf
```diff
@@ -30,29 +30,25 @@ variable "create_model" {
...
-  type        = string
-  default     = ""
-  description = "COS CA certificate"
+variable "cos" {
+  description = "Observability settings"
+  type = object({
+    model           = optional(string, "cos")
+    external        = optional(bool, false)
```
Batalex, line 39: **suggestion**: How about using some kind of 3 values enum here? Something like

```hcl
variable "cos_deployed" {
  type        = string
  description = "some test value"

  validation {
    condition     = contains(["external", "bundle", "no"], var.cos_deployed)
    error_message = "Valid values for var: test_variable are ("external", "bundle", "no")."
  } 
}
```
We can integrate that a little bit better in the `cos` variable so that we check that if set to "external" then we have the corresponding offers. In my test I just replace the bool `external` by a str `deployed` without validation.
That way, we can drop the local variable that was causing the issue:
```hcl
module "observability" {
  depends_on       = [module.spark, module.cos]
  count            = var.cos.deployed != "no" ? 1 : 0
  source           = "./modules/observability"
  dashboards_offer = var.cos.deployed == "external" ? var.cos.offers.dashboard : one(module.cos[*].dashboards_offer)
  logging_offer    = var.cos.deployed == "external" ? var.cos.offers.logging : one(module.cos[*].logging_offer)
  metrics_offer    = var.cos.deployed == "external" ? var.cos.offers.metrics : one(module.cos[*].metrics_offer)
  spark_model      = var.model
  spark_charms     = module.spark.charms
}

```

I gave it a try locally, and it seems to solve the "2 apply" issue:

**No COS**
```hcl
module "cos" {
  depends_on = [juju_model.cos]
  count      = 0
  source     = "../3.4/terraform/external/cos"
  model      = juju_model.cos.name
}

module "spark_deployment" {
  depends_on = [module.cos]
  source     = "../3.4/terraform"
  cos = {
    deployed = "no",
  }
  s3 = {
    bucket   = "spark-test",
    endpoint = "http://10.152.183.249:80"
  }
  create_model = true
}
output "charms" {
  description = "The name of the charms which are part of the deployment."
  value = merge(
    concat(module.spark_deployment[*].charms, module.cos[*].charms)...
  )
}
```
```
> tf apply
Plan: 17 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + charms = {
      + certificates   = "certificates"
      + history_server = "history-server"
      + hub            = "integration-hub"
      + kyuubi         = "kyuubi"
      + kyuubi_users   = "kyuubi-users"
      + metastore      = "metastore"
      + s3             = "s3"
      + zookeeper      = "zookeeper"
    }
```

**Bundled**
```hcl
module "spark_deployment" {
  depends_on = [module.cos]
  source     = "../3.4/terraform"
  cos = {
    deployed = "bundle",
  }
  s3 = {
    bucket   = "spark-test",
    endpoint = "http://10.152.183.249:80"
  }
  create_model = true
}
```
```
Plan: 65 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + charms = {
      + agent             = "agent"
      + alertmanager      = "alertmanager"
      + catalogue         = "catalogue"
      + certificates      = "certificates"
      + cos_configuration = "cos-configuration"
      + grafana           = "grafana"
      + history_server    = "history-server"
      + hub               = "integration-hub"
      + kyuubi            = "kyuubi"
      + kyuubi_users      = "kyuubi-users"
      + loki              = "loki"
      + metastore         = "metastore"
      + prometheus        = "prometheus"
      + pushgateway       = "pushgateway"
      + s3                = "s3"
      + scrape_config     = "scrape-config"
      + zookeeper         = "zookeeper"
    }
```

**External**
```hcl
module "cos" {
  depends_on = [juju_model.cos]
  source = "../3.4/terraform/external/cos"
  model  = juju_model.cos.name
}


module "spark_deployment" {
  depends_on = [module.cos]
  source     = "../3.4/terraform"
  cos = {
    deployed = "external",
    offers = {
      dashboard = module.cos.dashboards_offer,
      metrics   = module.cos.metrics_offer,
      logging   = module.cos.logging_offer
    }
  }
  s3 = {
    bucket   = "spark-test",
    endpoint = "http://10.152.183.249:80"
  }
  create_model = true
}
```
```
Plan: 64 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + charms = {
      + agent             = "agent"
      + alertmanager      = "alertmanager"
      + catalogue         = "catalogue"
      + certificates      = "certificates"
      + cos_configuration = "cos-configuration"
      + grafana           = "grafana"
      + history_server    = "history-server"
      + hub               = "integration-hub"
      + kyuubi            = "kyuubi"
      + kyuubi_users      = "kyuubi-users"
      + loki              = "loki"
      + metastore         = "metastore"
      + prometheus        = "prometheus"
      + pushgateway       = "pushgateway"
      + s3                = "s3"
      + scrape_config     = "scrape-config"
      + zookeeper         = "zookeeper"
    }
```

releases/3.4/terraform/variables.tf
deusebio: But if they are not used in a tf module, they should not be there. I suppose ManSol use these variables elsewhere, and not as input to our module, right?

Batalex: We have them in the `cos` module, but yeah, they are not wired up in our main.tf file. I believe we should keep them and use them in our main.tf file (and document them in the readmes)

releases/3.4/terraform/variables.tf
deusebio: Yes, I tried to see whether TF would be smart enough to create resources based on the type (whether something would be null or not), but it does not seem to be this smart. I believe I will go with your suggestion, it is also the most explicit and clear from a user standpoint

releases/3.4/terraform/variables.tf
deusebio: yes, I have wired them up in the main.tf in [c8be691](https://github.com/canonical/spark-k8s-bundle/pull/111/commits/c8be6919d8402f3928e3f6c60611836a5bef7249)

deusebio:
@Batalex I have updated the module with your suggestion. Also this allows to create some validation in the variable definition which makes things much clearer on a UX perspective. I very much like the new structure better now

deusebio:
One thing I'd also like to get thoughts is whether to use nested variables (like [here](https://github.com/canonical/spark-k8s-bundle/pull/111/files#diff-63560d57f0238784e84a16c1f384d587420365751d33ab1a516fb87872a7efd4R35)) or flattened ones using underscores instead. 

Maybe one reason to use the nested one is that we can encapsulate the validation logic on that particular variable. If we were to break it down into separate variable, it might be a bit harder to do validation across multiple variables (not sure though, I don't know TF enough)

releases/3.4/terraform/main.tf
```diff
@@ -51,6 +39,19 @@ module "spark" {
   zookeeper_units           = var.zookeeper_units
   tls_app_name              = module.ssc.app_name
   tls_certificates_endpoint = module.ssc.provides.certificates
+
+  spark_history_server_revision  = var.spark_history_server_revision != null ? var.spark_history_server_revision : local.revisions.spark_history_server
+  spark_history_server_image     = var.spark_history_server_image != null ? var.spark_history_server_image : local.images.spark_history_server
+  spark_integration_hub_revision = var.spark_integration_hub_revision != null ? var.spark_integration_hub_revision : local.revisions.spark_integration_hub
+  spark_integration_hub_image    = var.spark_integration_hub_image != null ? var.spark_integration_hub_image : local.images.spark_integration_hub
```
deusebio, line 46: **thought** I have deviated slightly from DA183, in the sense that the spec suggest to call these `spark_history_server_k8s_revision`, but I'm actually thinking if it would be better for conciseness to shorten this and remove at least `k8s` which seems really not needed. Actually, I'm even considering to just put `history_server_revision` and `integration_hub_revision`, without spark which is quite obvious

WDYT?

Batalex: I prefer the concise version as well. Using the full name is a demand from Mykola though

Batalex: From what I understand from this
> Can we use the charm name in cases where the only one function is possible?

The demand seems to be to not rename applications prefix for tf variables, just keep the charm name is there is only one of its kind

releases/3.4/terraform/main.tf
```diff
@@ -69,22 +72,35 @@ module "s3" {
...
-  metrics_offer    = one(module.cos[*].metrics_offer)
-  logging_offer    = one(module.cos[*].logging_offer)
+  dashboards_offer = var.cos.deployed == "external" ? var.cos.offers.dashboard : one(module.external_cos[*].dashboards_offer)
+  logging_offer    = var.cos.deployed == "external" ? var.cos.offers.logging : one(module.external_cos[*].logging_offer)
+  metrics_offer    = var.cos.deployed == "external" ? var.cos.offers.metrics : one(module.external_cos[*].metrics_offer)
   spark_model      = var.model
   spark_charms     = module.spark.charms
+
+  grafana_agent_revision = var.grafana_agent_revision != null ? var.grafana_agent_revision : local.revisions.grafana_agent
+  # grafana_agent_image = var.grafana_agent_image != null ? var.grafana_agent_image : local.images.grafana_agent
```
welpaolo, line 101: Do we want to have this flexibility ? I don't think so

releases/3.4/terraform/modules/observability/applications.tf
```diff
@@ -37,6 +38,7 @@ resource "juju_application" "pushgateway" {
   charm {
     name    = "prometheus-pushgateway-k8s"
     channel = "latest/stable"
```
welpaolo, line 40: channel 1/stable is available

releases/3.4/terraform/modules/observability/applications.tf
```diff
@@ -7,7 +7,7 @@ resource "juju_application" "agent" {
   charm {
     name     = "grafana-agent-k8s"
     channel  = "latest/stable"
```
welpaolo, line 9: channel 1/stable is available

releases/3.4/terraform/modules/observability/applications.tf
```diff
@@ -20,6 +20,7 @@ resource "juju_application" "cos_configuration" {
   charm {
     name    = "cos-configuration-k8s"
     channel = "latest/stable"
```
welpaolo, line 22: Same here

releases/3.4/terraform/modules/observability/applications.tf
```diff
@@ -51,7 +53,7 @@ resource "juju_application" "scrape_config" {
   charm {
     name     = "prometheus-scrape-config-k8s"
     channel  = "latest/stable"
```
welpaolo, line 55: Same here

releases/3.5/terraform/modules/observability/applications.tf
```diff
@@ -5,8 +5,9 @@ resource "juju_application" "agent" {
   name  = "agent"
   model = data.juju_model.spark.name
   charm {
-    name    = "grafana-agent-k8s"
-    channel = "latest/stable"
+    name     = "grafana-agent-k8s"
+    channel  = "latest/stable"
```
welpaolo, line 9: As above

releases/3.5/terraform/versions.tf
```diff
@@ -0,0 +1,51 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+terraform {
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = ">=0.16.0"
```
welpaolo, line 8:
```suggestion
      version = ">=0.19.0"
```

releases/3.4/terraform/versions.tf
```diff
@@ -0,0 +1,51 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+terraform {
+  required_providers {
+    juju = {
+      source  = "juju/juju"
+      version = ">=0.16.0"
```
welpaolo, line 8:
```suggestion
      version = ">=0.19.0"
```

welpaolo (approved):
Great work Enrico! I really like the flexibility provided by this update!

releases/3.4/terraform/modules/observability/applications.tf
deusebio: Setting the channel to stable is part of a different PR: https://github.com/canonical/spark-k8s-bundle/pull/113

I will merge this PR once that one lands, and rebasing on top of main

releases/3.4/terraform/modules/observability/applications.tf
deusebio: as per [this](https://github.com/canonical/spark-k8s-bundle/pull/111#discussion_r2137108345)

releases/3.4/terraform/versions.tf
deusebio: I don't think I'm using in this PR any particular functionality that requires TF provider to be above 0.19, and also 0.16 would work actually

Why do you think we need to pin to higher version? I'm just afraid that this may reduce our compatibility with other module (e.g. cos, tls, etc) to create an uber module.

welpaolo: Make sense!

releases/3.4/terraform/main.tf
deusebio: addressed in https://github.com/canonical/spark-k8s-bundle/pull/111/commits/5a98931daa4f36b4d1f13266a04e84a72b1c74aa

releases/3.4/terraform/main.tf
deusebio: grafana-agent is part of our applications, meaning that if we specify images for say integration-hub, why not doing this also for the other components to keep consistent. I would probably review this after v3, to see also if we want to remove all images all together

releases/3.4/terraform/main.tf
```diff
@@ -69,22 +72,35 @@ module "s3" {
...
+
+  s3_revision = var.s3_revision != null ? var.s3_revision : local.revisions.s3
 }
 
-module "cos" {
-  depends_on = [juju_model.cos]
-  count      = var.cos_model == null ? 0 : 1
-  source     = "./external/cos"
-  model      = var.cos_model
+module "external_cos" {
```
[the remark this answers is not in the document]
Batalex: Oh yes I remember that, but back then the `test.tf` file was from the POV of the external user, so the cos module defined there was the "external" one and we still had a bundled one, even if both were pointing to the same resource. Sorry for being confusing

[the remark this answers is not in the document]
Batalex: No, I don't think the name would change. Even if we point to a different repository, it's still a cos bundle that comes with our deployment, whereas in production users could use an already existing one not "managed/tracked" by our terraform file and relate to spark using cross model offers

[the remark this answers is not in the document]
deusebio: all right! Sounds good! Renaming!

releases/3.4/terraform/modules/observability/variables.tf
```diff
@@ -43,3 +43,39 @@ variable "logging_offer" {
...
   nullable    = false
 }
+
+variable grafana_agent_revision {
+  description = "Charm revision for grafana-agent-k8s"
+  type        = number
+  nullable    = false
+}
+
+# variable grafana_agent_image {
```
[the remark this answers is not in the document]
deusebio: Removed!

releases/3.4/terraform/modules/observability/variables.tf
```diff
@@ -43,3 +43,39 @@ variable "logging_offer" {
...
+  nullable    = false
+}
+
+variable pushgateway_revision {
+  description = "Charm revision for prometheus-pushgateway-k8s"
+  type        = number
+  nullable    = false
+}
+
+# variable pushgateway_image {
```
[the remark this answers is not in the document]
deusebio: Removed!

releases/3.5/terraform/README.md
```diff
@@ -40,15 +40,15 @@ module "ssc" {
 
 module "spark" {
   depends_on                = [juju_model.spark]
-  source                    = "git::https://github.com/canonical/spark-k8s-bundle//releases/3.5/terraform/modules/spark"
+  source                    = "./releases/3.5/terraform/modules/spark"
```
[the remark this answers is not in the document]
deusebio: Done!

releases/3.4/terraform/main.tf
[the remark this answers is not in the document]
deusebio: Done in [9cd4c50](https://github.com/canonical/spark-k8s-bundle/pull/111/commits/9cd4c5085adf416e10996324fb527515f3f6bc9c)

Batalex (approved):

PR #114, opened 2025-06-06
[DPE-7375] Backup and Restore tests for Kyuubi bundle and CI Improvements

theoctober19th:
The PR includes the following changes:
1) Adds Backup Restore tests in the CI
2) However, to do so, the existing fixture apparatus needed quite some refactoring -- which has been done in the same PR.
3) While struggling with the CI, I also realized that we have some regression since our Jubilant migration, where YAML tests are consistently failing -- however it is very very difficult to pinpoint that looking at some 100+ list of tests in the CI. To make it easier to understand which combination of tests failed in a quick look, I've also added some tooling that renders the results of the test matrix in a tabular format. For example:

image

I tried getting into the root cause of why YAML tests were failing -- most of them were pinpointed to COS bundle. However, once I started trying to fix them, I realized the PR exploded a lot in its scope and I had to draw a line somewhere. I've come to an conclusion to move forward backup-restore tests while coming back to the failing tests in Full Test Suite later in a dedicated PR.


python/tests/integration/conftest.py
```diff
@@ -645,3 +466,142 @@ def _forwarder(pod: str, port: int, namespace: str, on_port: int | None = None):
...
+
+@pytest.fixture(scope="class")
+def tempdir():
+    """A temporary directory to be used for tests."""
+    with local_tmp_folder() as tmp_folder:
+        yield tmp_folder
+
+
+@pytest.fixture(scope="class")
+def spark_bundle(
```
theoctober19th, line 479: I have made this and a bunch of ther fixtures "class" scoped such that the existing tests be left unaffected by simply wrapping them up with class, and for the backup -restore tests, we have the ability to teardown the bundle and reconstruct it in the middle of the test.

python/tests/integration/test_backup_restore.py
```diff
@@ -0,0 +1,689 @@
...
+    ).stdout.strip()
+
+    yield {
+        "endpoint": f"https://{host_ip}",
+        "access-key": key_id,
+        "secret-key": secret_key,
+        "tls-ca": ca_crt_base64,
+    }
+
+    subprocess.run(["sudo", "snap", "remove", "microceph", "--purge"], check=True)
```
Batalex, line 233: **suggestion**: Can we rework the microceph logic to avoid destructive operations while we don't have spread testing? This would be a little annoying to run locally

Batalex (commented):
Not yet done with the review, will revisit later

python/tests/integration/conftest.py
```diff
@@ -144,38 +144,41 @@ def pytest_addoption(parser):
...
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="class")
 def juju(request: pytest.FixtureRequest):
     keep_models = bool(request.config.getoption("--keep-models"))
     model = request.config.getoption("--model")
 
     if model is None:
         with jubilant.temp_model(keep=keep_models) as juju:
             juju.wait_timeout = 30 * 60
-            juju.model_config({"update-status-hook-interval": "60s"})
```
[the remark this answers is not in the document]
theoctober19th: sorry, this wasn't supposed to be deleted. I'll add this back.

python/tests/integration/conftest.py
```diff
@@ -431,123 +275,100 @@ def track_model(model: str) -> Generator[jubilant.Juju]:
     yield juju
 
 
-def spark_bundle_with_s3(juju: jubilant.Juju, credentials, bucket, bundle, cos):
-    """Deploy all applications in the Kyuubi bundle, wait for all of them to be active,
-    and finally yield a list of the names of the applications that were deployed.
-    """
+@contextlib.contextmanager
+def switch_model(juju: jubilant.Juju, model: str) -> Generator[jubilant.Juju]:
```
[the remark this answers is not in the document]
theoctober19th: The existing `track_model` did change the `juju.model` to the desired model and return the `juju` object to us, but it didn't actually switch the model to the new model (for example simulating exactly what `juju switch xxx` would do.)

The `juju offer` command is a bit weird in the sense that juju needs to be switched to the model where you make the offer -- and this isn't possible with just changing `model` parameter in the `juju` object. Therefore, the `switch_model` method exists for the purpose of doing `juju switch abc` and then during teardown `juju switch back-to-where-i-was`.

You're right regarding the docstring. I'll update it accordingly.

python/tests/integration/test_backup_restore.py
theoctober19th: I think we can add some condition to this fixture for example `if os.environ["CI"]` or something so that these run only on CI and not locally. I'll give that a try.

.github/scripts/aggregate_results.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
```
deusebio, line 1: **praise** this is a very very nice idea. Well thought!

.github/workflows/ci-tests-full.yaml
```diff
@@ -66,7 +81,43 @@ jobs:
...
-          juju-snap-channel: ${{ matrix.juju.snap_channel || '3.6/stable' }}
-          juju-agent-version: ${{ matrix.juju.agent || '3.6.0' }}
+          juju-snap-channel: ${{ matrix.juju-snap-channel }}
+          juju-agent-version: ${{ matrix.juju-agent-version }}
           azure-storage-account: ${{ secrets.AZURE_STORAGE_ACCOUNT }}
           azure-storage-key: ${{ secrets.AZURE_STORAGE_KEY }}
+          pytest-args: ${{ matrix.tox-env != 'integration-backup-restore' && '--keep-models' || '' }}
+
+
+  aggregate-results:
```
deusebio, line 91: **todo** remove and place on a different PR

deusebio (approved):
I don't have strong concerns with the work here, only a few suggestions and comments. Therefore, I'm fine to approve (and not re-review), but the only comment I'd like to follow up is that in general I agree with the main comment in the PR description of avoiding scope creep in this PR, and the only small change of the PR that I believe it would be beneficial would be to move the `aggregate_results` part on a different PR. I believe it is a very very sensible and good idea, it just does not belong to back and restore tests, and in the dedicated PR we can also discuss how to run the full set of tests

python/tests/integration/test_backup_restore.py
theoctober19th: We can indeed use an existing installation of microceph -- however the step `microceph enable rgw --ssl-certificate xxx --ssl-private-key xx` needs to be done during the test itself (because we would have a different SSL certificate and private key). So if the user already has enabled `rgw` once, that installation cannot be used unless it is disabled and enabled again with the different SSL certificate.

Batalex: Let's forget about this, we'll probably have spread testing at some point anyway

.github/workflows/ci-tests-full.yaml
```diff
@@ -6,6 +6,11 @@ name: Nightly CI tests
 on:
   schedule:
     - cron: "53 0 * * *" # Daily at 00:53 UTC
+  push:
+    tags:
+      # Force pushing this tag will trigger full set of tests
+      - test-run-full-test-suite
```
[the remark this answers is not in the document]
theoctober19th: This is a good idea, I've made this job a manual job.

python/tests/integration/conftest.py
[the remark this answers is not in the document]
theoctober19th: Added this back in the last commit.

python/tests/integration/conftest.py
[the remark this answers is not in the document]
theoctober19th: The `track_model` turned out to be something we can do on-the-fly. I've removed `track_model` and kept `switch_model` for now.

python/tests/integration/test_backup_restore.py
```diff
@@ -0,0 +1,689 @@
...
+        )
+
+    def test_reintegrate_kyuubi_with_metastore(self, juju: jubilant.Juju):
+        """Add backup kyuubi <> metastore relation after the metastore restoration is complete."""
+        # Now re-integrate the metastore back to kyuubi
+        juju.integrate(METASTORE_APP_NAME, f"{KYUUBI_APP_NAME}:metastore-db")
+        juju.wait(jubilant.all_active, delay=5)
+
+        # /opt/hive/bin/schematool -validate should return 0
+        output = juju.ssh(
```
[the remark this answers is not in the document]
theoctober19th: We already do this during relation creation as well us repeatedly during status computation as well. This exists here as an extra check -- but you're right, this check is just redundant that if validation failed, the charm would not be in active state in the first place, which we already check in the earlier lines.

.github/scripts/aggregate_results.py
theoctober19th: Sure, I'll try to rebase the changes into two different commits.

.github/workflows/ci-tests-full.yaml
theoctober19th: Sure, I'll rebase the changes into two different commits.

python/tests/integration/conftest.py
Batalex: Hum, I may have one last idea we could test, but I don't have strong opinions whether that's better for readability/maintainance.
We could create a new `tests/integration/fixtures_impl.py` file with the fixtures' body. Let's take the `namespace` fixture as an example.
Let's define `tests/integration/fixtures_impl.py:_namespace` as follows

```py
def namespace(namespace_name):
    return namespace_name
```

Yes, that's the same signature as what we currently have as a fixture in `tests/integration/conftest.py`.
Then, we replace the current fixture in  `tests/integration/conftest.py` by
```py
import pytest 
import .fixtures_impl 

namespace = pytest.fixture(scope="module")(fixtures_impl.namespace)
```

and in `test_backup_restore.py` we would just declare new fixtures using a different scope

```py
import pytest 
import .fixtures_impl 

namespace = pytest.fixture(scope="class")(fixtures_impl.namespace)
# rest of the file
```

We need to re-declare all fixtures in that file, even those which do not explicitely appear in the test function signatures, otherwise pytest would pick the module-scoped ones from `conftest.py`

python/spark_test/core/bundle/yaml.py
```diff
@@ -0,0 +1,83 @@
...
+import os
+from pathlib import Path
+
+import jinja2
+import yaml
+
+from . import BundleBackend, BundleBackendEnum
+
+
+class YamlBackend(BundleBackend):
```
Batalex, line 16: **praise**: I like those new abstractions, the cascading fixtures we had before were difficult to follow at first glance

python/tests/integration/test_backup_restore.py
```diff
@@ -0,0 +1,681 @@
+import json
```
Batalex, line 1: **praise**: Good test

Batalex (approved):
LGTM, but yeah, I would love for us to reduce the amount of diff here to keep as much historic context as possible in the git blame

Let's give https://github.com/canonical/spark-k8s-bundle/pull/114#discussion_r2156936551 a try

python/tests/integration/conftest.py
theoctober19th: TBH, I'm not keen on this idea too much because 1) this needs redeclaration of all fixtures in the file and 2) it feels a bit too hacky to my liking to use the decorator as a second order function here.

However, thank you so much for the pointer to the dynamic scope. I've found a way we can do this without having to expose config option to `pytest`. It seems the `config` object passed to `determine_scope` has access to the args that pytest was passed with.. and using it we can determine what test is calling this fixture. With the check if the test calling fixture is backup-restore test, we can return scope as "class" and for everything else, the scope could be module.

theoctober19th: Fixed in [7e6fc05](https://github.com/canonical/spark-k8s-bundle/pull/114/commits/7e6fc05a9b31d32504bdbd2e0446848545b6a2c1).

python/pyproject.toml
```diff
@@ -53,10 +53,10 @@ optional = true
 [tool.poetry.group.integration.dependencies]
 pytest-cov = ">=3.0"
 pytest-mock = ">=3.10"
-psycopg2-binary = "^2.9.9"
+psycopg2 = "^2.9.9"
```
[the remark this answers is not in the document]
theoctober19th: since we use this just for integration tests, I think we can keep the pre-built binary. I'll revert the change.

python/tests/integration/test_backup_restore.py
theoctober19th: Thanks!

welpaolo (approved):
Great work Bikalpa! Just left a minor comment

PR #116, opened 2025-06-10
[DPE-7414] create reference architecture

deusebio:
The successful run of the reference architecture is testified in the run here:

* https://github.com/canonical/spark-k8s-bundle/actions/runs/15764479975/job/44438061247?pr=116


python/tests/integration/test_kyuubi.py
```diff
@@ -70,11 +70,15 @@ def test_jdbc_endpoint(juju: jubilant.Juju) -> None:
...
 
-    client = KyuubiClient(**credentials, use_ssl=True, ca_cert=ca_cert)
+    # items = ast.literal_eval(task.results["certificates"])
+    # certificates = json.loads(items[0])
+    # ca_cert = certificates["ca"]
+
+    client = KyuubiClient(
+        **credentials,
+        use_ssl=True,
+    )  # ca_cert=ca_cert)
```
Batalex, line 81: **suggestion**: Should we add a fixture or something to disable ca cert verification? We definitely need to fix that to have a solution that works in all environment, but disabling it everywhere in the meantime might be bad as well

python/tests/reference_architecture/azure/main.tf
```diff
@@ -0,0 +1,71 @@
...
+module "cos" {
+  depends_on = [juju_model.cos]
+  source     = "git::https://github.com/canonical/spark-k8s-bundle//releases/3.4/terraform/external/cos"
+  model      = juju_model.cos.name
+}
+
+
+module "spark" {
+  # source                	= "git::https://github.com/canonical/spark-k8s-bundle//releases/3.4/terraform?ref=dpe-7336-enable-cos-deployment-configuration"
+  source                	= "../../../../releases/3.4/terraform"
```
Batalex, line 50: **thought**: I wish we had a `ROOT_DIR` env or something to always reference paths from the "project root"

deusebio: oh, now that we have merged the other PR, I can just not do the relative reference, but use the git reference properly

.github/workflows/deploy-to-aks.yaml
```diff
@@ -0,0 +1,142 @@
+name: Reference Architecture Testing (AKS)
+on:
+  pull_request:
+#   workflow_dispatch:
```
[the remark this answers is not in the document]
deusebio: Oh yes, I had this idea, but what do you think? If it makes sense to parametrize this, I can work to break this into a "dispatch workflow" and then a CI where we input some values. I believe it would be a sensible structure, such that we can also run this tests against other versions on demand

python/tests/integration/test_kyuubi.py
```diff
@@ -70,11 +70,15 @@ def test_jdbc_endpoint(juju: jubilant.Juju) -> None:
     self_signed_certificate_unit = next(iter(status.apps["certificates"].units.keys()))
     task = juju.run(self_signed_certificate_unit, "get-issued-certificates")
     assert task.return_code == 0
-    items = ast.literal_eval(task.results["certificates"])
-    certificates = json.loads(items[0])
-    ca_cert = certificates["ca"]
 
-    client = KyuubiClient(**credentials, use_ssl=True, ca_cert=ca_cert)
+    # items = ast.literal_eval(task.results["certificates"])
```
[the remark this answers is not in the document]
deusebio: I have removed this because the certificates have the `kyuubi-0` common name which does not correspond to the endpoint, and the CA verification fails. It related to the point of Alex here: https://github.com/canonical/spark-k8s-bundle/pull/116#discussion_r2159114292. Please provide your thoughts on this, whether you would prefer to add a fixture, remove it or anything else. 

I don't have strong ideas and I'm fine with whatever the majority prefer to go for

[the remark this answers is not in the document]
theoctober19th: [This PR](https://github.com/canonical/kyuubi-k8s-operator/pull/96) once merged will fix this in Kyuubi as it tries to set the same endpoint that Data Integrator uses as CN of the server certificate. As for when Kyuubi is not exposed via loadbalancer, we can use [this](https://github.com/canonical/kyuubi-k8s-operator/pull/96/files#diff-e9349e06331f9876f984bcf05671235cfbe546046bccab6ad1abb0b183982b30R700) context manager to temporarily patch kyuubi's internal DNS name to its pod IP address in the tests.

For cases where the CN name is > 64 characters, we can't really do anything in the CN than to just return first 64 characters and then hope the client is able to read SANs field instead of relying on CN.

python/tests/integration/test_kyuubi.py
theoctober19th: I agree with @deusebio here. We need to fix this issue. Actually, I'm kinda hopeful [this PR](https://github.com/canonical/kyuubi-k8s-operator/pull/96) will fix this. I'm currently trying to sync it with the bundle and see if the existing tests pass.

.github/workflows/deploy-to-aks.yaml
```diff
@@ -0,0 +1,142 @@
...
+          # Renaming of the version to use _ instead of . 
+          RUN_VERSION="spark-${_K8S_VERSION//./_}"
+          RESOURCE_GROUP=${RUN_VERSION}_rg
+          NAME=${RUN_VERSION}_aks
+          LOCATION=westeurope
+          echo "RESOURCE_GROUP=${RESOURCE_GROUP}" >> $GITHUB_ENV
+          echo "NAME=${NAME}" >> $GITHUB_ENV
+          echo "LOCATION=${LOCATION}" >> $GITHUB_ENV
+          az group create --name ${RESOURCE_GROUP} --location ${LOCATION}
+          az aks create \
```
theoctober19th, line 70: I have a few questions around here regarding the way we're bootstraping the AKS cluster. Since we're using the Terraform module later anyway, why not simply put all this bootstrapping logic in the TF itself? We can also maybe separate the cluster bootstrap and the bundle bootstrap Terraform separately. I believe it would be more readable as well as more reproducible to have it in TF than bash here.

theoctober19th (commented):
I like the idea of having the tests for the reference architecture, something that was playing in the back of my mind for some time.. However, I have some questions regarding how we can better bootstrap the cluster for the tests to run.

.github/workflows/deploy-to-aks.yaml
deusebio: Actually, I was playing with this and trying this out, and I DEFINITELY agree with you that this would be *nicer* to have everything in TF. however it is not as easy as it seems, so I gave up because it was a bit outside of the scope and taking longer that I had reserved for. 

Let me explain why:
1. You can't have everything in one single TF module, because TF JUju provider assumes you already have a controller that is bootstrapped. 
2. Setting up the infrastructure **only** does not make much sense, if you then still use bash to setup the controller, such that you can just have two tf modules, one for the infra/juju and one for the deployment. So I was looking into doing everything (setting up infra + juju controller) in one go.
3. I was playing with the TF module you [had worked with](https://github.com/theoctober19th/aks-validation), but there were some hiccups with the bootstrap process from time to time, I had to investigate
4. Those TF modules still requires to export the various controller informations (endpoints, username, password, ca, etc) as outputs

So I ended up considering that it was a nice-to-have, and a lot more work than just using the bash commands (`az aks` + `juju bootstrap`)

python/tests/reference_architecture/azure/main.tf
deusebio: Done

theoctober19th (approved):
LGTM. I like the idea for having the workflow as parameterized manual job.

PR #121, opened 2025-06-30
[DPE-7701] Upgrade bundle to remove Kyuubi charm actions and incorporate TLS lib upgrade (#121)

theoctober19th:
This PR:
* Upgrades the version / revision of `kyuubi-k8s` in the Spark Bundle so that the recent changes in removal of `get/set-password` and `set-tls-private-key` actions are reflected.
* Expose admin password and custom tls private key as variables in the TF bundle (and make them optional to comply with the behavior in the charm where these are optional)
* Update relevant integration tests to comply with the newer syntax of supplying admin password and private key
* Pin `self-signed-certificates` to `rev324` temporarily (since `self-signed-certificates` recently bumped its provider requirements to `juju >= 0.20` and we're not ready to upgrade right away (at least in this PR because we don't want the scope to mutliply)
* Pin the `juju` provider to `<0.20` because v0.20 introduces some breaking changes which we want to resolve in a separate PR.
* Include the recent changes in Kyuubi regarding the config option `profile`.
* Upgrade integration hub to latest revision 58.


Batalex (changes requested):
Overall LGTM, but one request

> Pin the juju provider to <0.20 because v0.20 introduces some breaking changes which we want to resolve in a separate PR.

I can only see the pin for 3.4 on the external cos module. I think this is an issue because 3.5 does not have the pin, neither does 3.4 if we deploy without cos. How about moving this pin to 3.{4,5}/terraform/versions.tf?

theoctober19th:
@Batalex The PR pins juju < 0.20 in  both 3.4 / 3.4 external COS, so I couldn't understand your comment. Could you please elaborate?

Batalex (approved):
Since the whole dependency chain is evaluated on `terraform init`, my inital comment is no longer blocking, just a suggestion.
Yes, I meant the uber module

.github/workflows/ci-tests-minimal.yaml
```diff
@@ -19,9 +19,9 @@ jobs:
       fail-fast: false
       matrix:
         tox-env:
-          - integration-sparkjob
-          - integration-bundle
-          - integration-kyuubi
+          # - integration-sparkjob
+          # - integration-bundle
+          # - integration-kyuubi
```
deusebio, line 24: **question,todo** should we re-enable these? Or why are we commenting them out?

releases/3.4/terraform/modules/spark/variables.tf
```diff
@@ -9,6 +9,27 @@ variable "model" {
   nullable    = false
 }
 
+variable "admin_password" {
+  description = "The password for the admin user."
+  type        = string
+  sensitive   = true
+  default     = null
```
deusebio, line 16: **question** what happens if this is null? If the config is not set, what is the password and how it can be retrieved?

python/tests/integration/test_backup_restore.py
```diff
@@ -247,6 +247,21 @@ def test_ha_enabled(self, juju: jubilant.Juju) -> None:
...
+        credentials = get_kyuubi_credentials(juju)
+        credentials.update({"username": "admin", "password": admin_password})
+        ca_cert = get_kyuubi_ca_cert(juju, certificates_app_name="certificates")
+        kyuubi_client = KyuubiClient(**credentials, use_ssl=True, ca_cert=ca_cert)
+
+        assert "default" in kyuubi_client.databases
+
+        # Store the old admin password into context, later we test whether
+        # the same password got restored for the admin user or not
+        context["old_admin_password"] = admin_password
```
deusebio, line 263: **question** I'm not sure why this is needed? Checking from the code, it would seen that both tests (here and when restoring) would always use the same admin password anyway, which I suppose is the fixture `adminpassword`

are we also testing password rotation?

deusebio (approved):
Looks good! I only have a few questions I'd like to get some answers for my own sake and understanding, but nothing really blocking

releases/3.4/terraform/main.tf
```diff
@@ -21,7 +21,7 @@ resource "juju_model" "cos" {
 
 module "ssc" {
   depends_on  = [juju_model.spark]
-  source      = "git::https://github.com/canonical/self-signed-certificates-operator//terraform"
+  source      = "git::https://github.com/canonical/self-signed-certificates-operator//terraform?ref=rev324"
```
deusebio, line 24: **todo** this change is missing on 3.5 version

theoctober19th: Fixed in commit cd6dd31.

releases/3.4/terraform/modules/spark/variables.tf
theoctober19th: If this config is null, the corresponding config `system-users` in the Kyuubi charm is not set. When this config in the charm is not set, the charm auto-generates the password for the admin user on its own, however this password can't be retrieved. If the user wants to set a custom password for admin user, they are required to provide this config option.

python/tests/integration/test_backup_restore.py
theoctober19th: I've added this test here so as to make sure we check the old admin password has been restored or not. This might be very trivial here because we basically deploy the TF with same variable twice.. however from a theoretical point of view (and from POV of the bundle, not the POV of how we deploy this), this test is still important. Even if we used the same variable during deployment, the assertion that this is picked correctly by the charm doesn't harm much. WDYT?

.github/workflows/ci-tests-minimal.yaml
theoctober19th: I've reverted this back to normal.

PR #125, opened 2025-07-15
[DPE-7719] Improve CI stability and add debug info

Batalex:
## Changes

- All changes from #124
- Add some debug information in case of test failure: disk usage, juju status, pods status
- Switched from GH runners to xlarge self hosted for integration test. Way more stable in my experience while deploying the full spark + cos ecosystems
- Bumped integration-hub to rev 64, the premature data access was a common cause of instability

### COS lite
I mentioned during daily that I would try the official cos-lite TF module instead of our own. In the end I did not, because
1. There is no tag at the moment, and I did not feel like pinning a commit hash on our side
2. Seems like they are on the verge on switching grafana-agent to a new otelcol charm. I don't want us to get a bad surprise all of a sudden

### CI Runner

I managed to clear 17GB of additional disk space by removing the following from the GH runner
```
        sudo rm -rf \
          /opt/google \
          /opt/hostedtoolcache \
          /opt/microsoft/powershell \
          /opt/microsoft/msedge \
          "$AGENT_TOOLSDIRECTORY" \
          /usr/lib/firefox \
          /usr/lib/mono \
          /usr/local/julia* \
          /usr/local/lib/android \
          /usr/local/share/chromium \
          /usr/local/share/powershell \
          /usr/share/dotnet \
          /usr/share/gradle* \
          /usr/share/miniconda \
          /usr/share/sbt \
          /usr/share/swift \
          /home/linuxbrew
```
but this did not improve the stability that much. We did have some disk pressure-related issues once in a while, but the main cause seems to be something else. I just wanted to write this here for future reference, if we or someone else have an issue with disk space on GH runner.
CPU seem to be 100% maxxed out during the total run, that could be an issue. We have 15GB of memory, that could also be the root cause.

Using `large` runner only led to a marginal stability improvement; this is why I settled on `xlarge`

### DIFF
--- a/.github/workflows/ci-uat.yaml
+++ b/.github/workflows/ci-uat.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   integration-tests:
     name: Integration Tests
-    runs-on: ["self-hosted", "linux", "X64", "jammy", "large"]
+    runs-on: ["self-hosted", "linux", "X64", "jammy", "xlarge"]
     timeout-minutes: 120
     strategy:
       fail-fast: false


python/tests/integration/conftest.py
```diff
@@ -616,21 +616,17 @@ def spark_bundle(
         cos_juju_model.model = cos
         logger.info("Waiting for COS deployment to settle down")
         cos_juju_model.wait(
-            ready=lambda status: jubilant.all_agents_idle(status, *COS_APPS),
-            error=lambda status: jubilant.any_error(status, *COS_APPS),
-            timeout=1800,
+            lambda status: jubilant.all_active(status, *COS_APPS),
```
theoctober19th, line 619: nit: I'd still prefer to put an error condition here that if any charm go to "error" state, we abort and report failure immediately. This would also help us catch errors in COS or other charms that we maintain early (and thus minimizing the risk that those errors go unnoticed just because a subsequent hook puts the charm back to good state).

python/tests/integration/conftest.py
```diff
@@ -616,21 +616,17 @@ def spark_bundle(
...
+            timeout=2400,  # 40 min
             delay=10,
         )
 
     logger.info("Waiting for spark deployment to settle down")
     juju.wait(
-        ready=lambda status: jubilant.all_active(
+        lambda status: jubilant.all_active(
```
theoctober19th, line 626: same as above.

releases/3.4/terraform/main.tf
```diff
@@ -21,7 +21,7 @@ resource "juju_model" "cos" {
 
 module "ssc" {
   depends_on  = [juju_model.spark]
-  source      = "git::https://github.com/canonical/self-signed-certificates-operator//terraform?ref=rev324"
+  source      = "git::https://github.com/canonical/self-signed-certificates-operator//terraform?ref=rev326"
```
theoctober19th, line 24: We can safely remove this pinning now.

theoctober19th (approved):
Great work! The results are impressive. And this combined with the other PR https://github.com/canonical/spark-k8s-bundle/pull/126 should get our CI into a lot more stable results.

python/tests/integration/conftest.py
Batalex: Hoping that the revision bump on cos-config will let us remove this soon!

releases/3.4/terraform/main.tf
theoctober19th: Nah, let's not go through the CI re-runs again, these will definitely be added with my PR.

PR #136, opened 2025-07-30
[DPE-7722] Re-enable Reference Architecture (AKS) testing 

deusebio:


theoctober19th (approved):
From what I understand here, `terraform destroy -auto-approve` can sometime unfortunately fail, but we can't do nothing about it than to go check for logs why it failed.

What is generally needed to be done in order to fix a broken destruction? Do all of us have access to these resources in the backend that we can log into Azure cloud and force-delete the resources by ourself?

deusebio:
> What is generally needed to be done in order to fix a broken destruction? Do all of us have access to these resources in the backend that we can log into Azure cloud and force-delete the resources by ourself?

I don't think there is much to worry about since what we are failing to destroy is only the Juju model. Tearing down the AKS cluster is done in the following step, which is carried out either way. So effectively, even before, if destroying the juju model would fail, resource would still be cleaned up, but the CI would appear as "red"

welpaolo (approved):
LGTM

PR #138, opened 2025-08-08
[MISC] Add `spark-service-account` and `metastore` offers in 3.5 terraform bundle

amc94:
terraform validate and apply for 3.5 release were failing because of missing offers. Copied ones from 3.4 release


deusebio (approved):
Looks good! Thank you for this! 

The two red tests are related to missing secrets in your repo which is required to run those tests. They will be fine in the main branch. So good to go!

PR #144, opened 2025-09-02
[MISC] Add missing offers in Spark 3.5

theoctober19th:


releases/3.5/terraform/modules/spark/offers.tf
```diff
@@ -0,0 +1,11 @@
...
+  model            = data.juju_model.spark.name
+  application_name = juju_application.integration_hub.name
+  endpoints        = ["spark-service-account"]
+}
+
+resource "juju_offer" "metastore" {
+  model            = data.juju_model.spark.name
+  application_name = juju_application.metastore.name
+  endpoints        = ["database"]
+}
```
Batalex, line 11: **question**: I understand why we would want to make an offer for the integration hub, for creating new service accounts, but what is the use case for the metastore? Do we actually need it here (and in 3.4)?

Batalex: Oh yes, that makes perfect sense thanks

welpaolo (approved):

PR #142, opened 2025-08-19
[DPE-7943] Improve stability of UAT and CI tests

theoctober19th:
The following reasons were identified for our constantly failing CI in UAT tests:

1) https://github.com/canonical/cos-configuration-k8s-operator/issues/128 This issue exists in revisions > 69. As workaround, we previously pinned to rev69.
2) However, https://github.com/canonical/cos-configuration-k8s-operator/issues/84 this exists in revisions > 65. As a workaround, I had to pin to rev65. I've opened a PR to fix this myself here: https://github.com/canonical/cos-configuration-k8s-operator/pull/131, however even after this is merged, we can't bump to latest revision until we fix https://github.com/canonical/cos-configuration-k8s-operator/issues/128.

PR #156, opened 2025-09-23
[MISC] Pin cos-configuration-k8s to rev65 on YAML bundle

theoctober19th:
Earlier [we pinned this in Terraform bundle](https://github.com/canonical/spark-k8s-bundle/pull/142), but since we have 1 test involving YAML, that was failing recently. To improve stability, now we want to pin this in YAML bundle as well.

Revision 65 is pinned because of the following two issues in `cos-configuration-k8s`:
 https://github.com/canonical/cos-configuration-k8s-operator/issues/128 
 https://github.com/canonical/cos-configuration-k8s-operator/issues/84


deusebio (approved):
LGTM!

welpaolo (approved):

PR #160, opened 2025-09-23
docs: Add redirects for RTD docs

izmalk:
Enabled rediraffe extension.
Added redirects from old paths and filenames to the new ones, on RTD.


izmalk:
> Looks good, but is there a way to test this?

Yes, there is a preview link in the checks section. You can select a random address from the `redirects.txt` and paste it after the PR preview link.

PR #161, opened 2025-09-26
chore: Add a pull request template with a checklist

izmalk:
## Description

Basic PR template with a checklist for docs ~~and cloud resources~~.

## Checklist

~~- [x] I have cleaned any remaining cloud resources from my accounts.~~

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,4 +5,3 @@
 ## Checklist
 
 - [ ] I have added or updated any relevant documentation.
-- [ ] I have cleaned any remaining cloud resources from my accounts.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,8 @@
+## Description
+
+<!-- add description here -->
+
+## Checklist
+
+- [ ] I have added or updated any relevant documentation.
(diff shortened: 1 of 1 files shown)


welpaolo (approved):
LGTM. I will let the final word to @deusebio :)

deusebio:
> I have added or updated any relevant documentation.

I'm fine as a reminder but often we break the documentation into separate tickets. So, overall, I don't think it is appropriate to require this in any PR. But I suppose one can always remove this when opening the PR, and this would always provide a good nudge to at least think about this. 

> I have cleaned any remaining cloud resources from my accounts.

I would instead frankly remove this. It is fine for the Jira ticket that measures the progress/operations, but I don't think it is relevant in a code PR on github. If someone from community would be opening the PR, what does this mean? I would skip this one honestly

deusebio (changes requested):
I would remove the cloud resources one

izmalk:
Yes, this is a non-mandatory reminder. It can be easily removed in a PR.
The docs one is a part of the Docs maintenance policy effort, described in the Documentation Dashboard as a quality objective.
The cloud resources one is an initiative from Mykola (I think it's a part of a bigger effort to reduce cloud costs), that we've applied in Kafka. It's up to you to decide, so I've removed it from here.

PR #167, opened 2025-10-27
[PRA-52] Pin tf provider to <1.0.0

Batalex:
Juju Terraform provider 1.0.0 just came out and has some [breaking changes](https://github.com/juju/terraform-provider-juju/releases/tag/v1.0.0).

We are not ready yet to use it for the bundle, as we rely on other modules which have yet to migrate to 1.0.0


theoctober19th (approved):
LGTM

PR #198, opened 2026-03-03
[PRA-210][PRA-211] Add advanced scheduling how-to documentation

Batalex:
##  Changes

This PR adds the how-to documentation presenting various strategies to schedule Charmed Apache Spark and its user-driven workloads.
Everything has been successfully tested on PS7 as part of the previous tickets on setting up a mixed architecture deployment.

At first I went for two different pages, one for the jobs themselves and one for the components, but I eventually regrouped the two in one place. The components' section is more limited than its counterpart and quite short, and we could argue that tainting the nodes dedicated to running user-driven workloads and configuring the Spark jobs to run on them takes care of the two original issues in one fell swoop. Therefore, I thought it made sense to have everything in one place considering this common thread.

I'm not comfortable presenting the 2-steps deployment using the namespace node affinity operator and its "necessary evil bug" in our documentation, so I have not included it yet. If requested, I can add it though.

See the live docs here: https://canonical-charmed-spark--198.com.readthedocs.build/198/how-to/advanced-scheduling/


deusebio (changes requested):
Looks good, and I believe the guide is very clear and direct to the point, that I really like it, and nicely fit into a how-to guide style. 

I see two improvements before approving:
1. In the userworkload maybe nudge to a pattern where one uses namespace-node-affity operator for **defaults** settings, while using pod-defaults for more fine-grained settings. It is true that one could also use pod-defaults with integration hub for default/centralized setting, but i believe we can probably rather nudge towards namespace-node-affinity operator, since using integration hub is a bit less represented as a code, since it relies on uploading a file to the object storage, whereas the node-affinity-operator may actually store this information more in the model, and it seems a more scalable approach

2. In the charm part, I believe we are missing the use case of deploying in nodes that have been tainted, which is a use-case that we may have, and here also nudging towards using the node-affinity-operator can be the way to go, as we also discuss this in the context of kubeflow, making these two presentation consistent one another.

welpaolo (commented):
Great work Alex! I believe we can merge after Enrico's comment are resolved

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+A single Juju model can contain applications deployed over different architectures.
+
+```{note}
+You may check if a charm supports a specific architecture on [Charmhub](https://charmhub.io/).
+```
+
+Constraints tags may also be used to set affinity/anti-affinity of the charms' pods.
+Please note that they are no native Juju mechanisms for setting tolerations, so the deployments examples here are limited to **untainted** nodes.
```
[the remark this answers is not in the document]
deusebio: > This does not cover the modeloperator pod and the namespace node affinity operator itself (cf. PR description).

Yes, sure! I believe it is fair. Unfortunately we cannot do much with that

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
+---
+myst:
+  html_meta:
+    description: "How-to guide for scheduling Charmed Apache Spark's components and its jobs on Kubernetes clusters."
```
izmalk, line 4: Praise: Kudos for adding the metadata description.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+---
+
+(how-to-advanced-scheduling)=
+
+# Advanced scheduling
+
+You can optimize infrastructure governance and performance by configuring Charmed Apache Spark with Kubernetes mechanisms, such as node affinity and toleration.
```
izmalk: Suggestion: I like the idea, but let's make it a bit more imperative. What I mean is the use of "can", which suggests the potential but does not imply the action. 


```suggestion
This guide shows how to configure Charmed Apache Spark with Kubernetes mechanisms such as node affinity and toleration to optimize infrastructure governance and performance.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+Those mechanisms are used to decouple control plane operations from user-driven workloads, ensuring system services remain stable on cost-effective instance.
+Spark executors can benefit from specialized hardware (high-memory nodes, custom hardware resources such as GPU, specific architecture) while maintaining the flexibility to scale idle resources to zero.
```
izmalk: Nitpick: I think we need to capitalise the Executors here, since it's being used with Spark in front. Just the way we mention it in out terminology guidance - https://github.com/canonical/spark-k8s-bundle/blob/main/python/CONTRIBUTING.md#terminology

We can use driver and executor later without capitalisation if we don't add Spark in front. It's weird, I know.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+## Prerequisites
+
+This section assumes that Charmed Apache Spark is deployed on a multi-nodes Kubernetes cluster.
+Advanced scheduling is achieved using [nodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector), [Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) and [Taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
+
+`nodeSelector` and affinities are notably applied to Kubernetes labels while taints are applied to the nodes themselves.
```
izmalk: Suggestion: Wording improvements.

```suggestion
This guide assumes that Charmed Apache Spark is deployed on a multi-node Kubernetes cluster.

To configure advanced scheduling, use [nodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector), [affinity rules](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity), and [taints](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
`nodeSelector` and affinity rules work with Kubernetes labels, while taints are applied directly to nodes.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+You may apply a label to a node using the following command:
```
izmalk: Todo: Such instructions should be much more imperative and calling for action, We can also make them much shorter and clearer by dropping unnecessary bits like "using the following".

```suggestion
To apply a label to a node:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+```shell
+kubectl label nodes <node_name> <key>=<value>
+```
+
+Tainting a node is achieved using the following:
```
izmalk: Suggestion: Apply the same format for all instructions if possible.

```suggestion
To taint a node:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+We recommend exclusively using `NoSchedule` effects rather than `NoExecute` to avoid disrupting pre-existing workloads.
+```
+
+Finally, you may use
+
+```shell
+kubectl describe node <node_name>
+```
+
+to verify that taints and labels were properly applied to the node.
```
izmalk: Suggestion: Let's put the action we are trying to do in front of the command for consistency and awareness. See suggestion above.


```suggestion
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+```shell
+kubectl taint nodes <node_name> <key>=<value>:<effect>
+```
+
+```{note}
+We recommend exclusively using `NoSchedule` effects rather than `NoExecute` to avoid disrupting pre-existing workloads.
+```
+
+Finally, you may use
```
izmalk:
```suggestion
Verify that taints and labels were properly applied to the node:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+```shell
+kubectl describe node <node_name>
+```
+
+to verify that taints and labels were properly applied to the node.
+
+This guide will demonstrate how to allow pods to be scheduled on tainted nodes and how to assign them to those nodes, ensuring that:
```
izmalk: Suggestion: Wording improvement. I'd like to avoid using future tense like this.

```suggestion
This guide shows how to schedule pods on tainted nodes and assign them to specific nodes, ensuring that:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+- Spark driver and executor pods are scheduled on the nodes dedicated to running user workloads
```
izmalk: Chore: The [terminology guide](https://github.com/canonical/spark-k8s-bundle/blob/main/python/CONTRIBUTING.md#terminology) dictates having both 'Driver' and 'Executor' capitalised. Apply everywhere.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+## Scheduling jobs
+
+Charmed Apache Spark can be configured to schedule driver and executor pod on specific nodes.
+This section details how to set up and configure the advanced scheduling of Spark jobs to mutually segregate control-plane workloads from user workloads and allocate pods on mixed architectures clusters.
+
+### Deploying the Namespace Node Affinity Operator (recommended)
+
+The [Namespace Node Affinity Operator](https://github.com/canonical/namespace-node-affinity-operator) adds a given set of node affinities and/or tolerations to all pods deployed in a namespace.
+This is a great match for Charmed Apache Spark, since we recommend running user-driven workloads in dedicated namespace rather than in the Charmed Apache Spark Juju model's.\
```
izmalk: Suggestion: A non-blocking wording improvement. 

```suggestion
This works well with Charmed Apache Spark, because user workloads are best run in a dedicated namespace rather than in the namespace used by the Charmed Apache Spark Juju model.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+To deploy the Namespace Node Affinity Operator, run the following command:
+
+```shell
+juju deploy -m <charmed_spark_juju_model> namespace-node-affinity --trust
```
izmalk: Question: I'm sorry, I'm confused here a bit. What is the model we are talking about here? It usually does not require one. We have models for History Server and Kyuubi, but not Spark itself, because we don't do `juju deploy spark`. That applies to the text above too.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+By default, the WebHook is not configured to modify pods in any namespace.
```
izmalk: Suggestion: I have doubts about the webhook capitalisation here. Let me know if it's correct somehow or fix it everywhere otherwise.


```suggestion
By default, the webhook is not configured to modify pods in any namespace.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+First, you must label all namespaces that will contain the pods you want to change the affinities/tolerations using:
```
izmalk: Suggestion: Wording improvement.

```suggestion
First, label each namespace that contains pods you want the webhook to modify by applying node affinity and toleration settings:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+Repeat for all namespaces which will contain Spark service accounts.
+
+```{note}
+Note that those namespaces need to exist before you label them.
+```
+
+Then, you must also pass the WebHook configuration to be applied to the pods.
+The example below will apply a `nodeSelector` to the pods in the namespace to assign them to a node with the matching label and a toleration to run on said tainted node.
```
izmalk: Suggestion: Wording improvement to make it more imperative and clear.

```suggestion
Then, provide the webhook configuration to apply to pods in those namespaces.
The following example applies a `nodeSelector` to pods in the namespace so that they are scheduled onto nodes with the specified label. It also adds a toleration so that the pods can run on nodes with the specified taint.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+  tolerations:
+  - key: <taint_key>
+    operator: Equal
+    value: <taint_value>
+    effect: <taint_effect>
+<namespace_2>: |
+  ...
+```
+
+You may save this file under `namespaces_settings.yaml` and configure the charm using:
```
izmalk: Suggestion: Wording improvement for imperative language.

```suggestion
Save this file as `namespaces_settings.yaml`, then configure the charm:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+```shell
+juju config -m <charmed_spark_juju_model> namespace-node-affinity settings_yaml="$(<namespaces_settings.yaml)"
+```
+
+This is it, you may now run a Spark job using the `spark-client` snap and see that the driver and executor pods are scheduled on the node(s) with the matching label, while the taint prevents other workloads from being scheduled there.
```
izmalk: Suggestion: Wording improvement for imperative language.

```suggestion
To verify the configuration, run a Spark job using the `spark-client` snap. The driver and executor pods are scheduled on nodes with the matching label, while the taint prevents other workloads from being scheduled on those nodes.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+  nodeSelectorTerms:
+    - matchExpressions:
+      - key: kubernetes.io/arch
+        operator: In
+        values:
+        - arm64
+```
+
+```{note}
+Charmed Apache Spark provides multi-architecture manifests for rock images supporting amd64 and arm64.
```
izmalk: Chore: Literals syntax.

```suggestion
Charmed Apache Spark provides multi-architecture manifests for rock images supporting `amd64` and `arm64`.
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+Therefore, your container runtime should be able to use the proper image based on the node architecture without any further intervention.
+```
+
+Two Namespace Node Affinity Operator applications can work alongside one another to enforce distinct configurations on the driver and executor pods.
+The solution is to use the `excludedLabels` to avoid applying the driver pod configuration to the executor pods and vice-versa.
+Here is how to apply a toleration to the non-driver pods on the first Namespace Node Affinity Operator:
```
izmalk:
```suggestion
To apply a toleration to the non-driver pods on the first Namespace Node Affinity Operator:
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+  tolerations:
+  - key: <taint_key>
+    operator: Equal
+    value: <taint_value>
+    effect: <taint_effect>
+```
+
+Respectively, doing the same thing on the second one by excluding `spark-role: executor` will result in a configuration applied to non-executor pods.
+
+### Defining a Pod template (alternative)
```
izmalk: Suggestion: Just an idea, up to you.

```suggestion
###  (Alternative) Defining a Pod template
```

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
+---
+myst:
+  html_meta:
+    description: "How-to guide for scheduling Charmed Apache Spark's components and its jobs on Kubernetes clusters."
+---
+
+(how-to-advanced-scheduling)=
+
+# Advanced scheduling
```
izmalk: Suggestion: Make it sound like a how-to guide, not a concept.

```suggestion
# How to use advanced scheduling
```

docs/how-to/index.md
```diff
@@ -78,5 +79,6 @@ spark-history-server/index
 Use K8s pods<use-k8s-pods.md>
 Streaming Jobs<streaming-jobs.md>
 Use GPU<use-gpu.md>
+Advanced Scheduling<advanced-scheduling.md>
```
izmalk, line 82: Note: We can keep the shorter name here for the Menu.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+
+- Spark driver and executor pods are scheduled on the nodes dedicated to running user workloads
+- Charmed Apache Spark components are scheduled on control-plane nodes and/or specific architecture
+
+## Scheduling jobs
+
+Charmed Apache Spark can be configured to schedule driver and executor pod on specific nodes.
+This section details how to set up and configure the advanced scheduling of Spark jobs to mutually segregate control-plane workloads from user workloads and allocate pods on mixed architectures clusters.
+
+### Deploying the Namespace Node Affinity Operator (recommended)
```
izmalk: Suggestion: I don't mind using gerund for headings in general. But we need to do it consistently. It seems OK on this page consistency-wise, but in other how-tos in Spark docs we mostly avoid gerund to make it sound more imperative. Do you think we can switch to non-gerund headings in this guide too? It's a non-blocking recommendation.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+## Scheduling Charmed Apache Spark components
+
+While the two previous sections already take care of segregating control-plane workloads from user-driven workloads, this guide details a few strategies on how to also separate the Charmed Apache Spark component from third party workloads (neither Charmed Apache Spark nor the Spark jobs) and take advantage of specific architectures.
+
+### Using Juju commands
+
+To target a specific architecture, a Juju constraint can be applied to the Charmed Apache Spark model itself, or to each individual charm.
+To apply the constraint to the model, run:
```
izmalk, line 215: Praise: Good imperative wording here.

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,339 @@
...
+Constraints tags may also be used to set affinity/anti-affinity of the charms' pods.
+Please note that they are no native Juju mechanisms for setting tolerations, so the deployments examples in this section are limited to **untainted** nodes.
+
+The following command:
+
+```shell
+juju deploy -m <charmed_spark_juju_model> kyuubi-k8s --constraints "tags=<label_key>=<label_value>"
+```
+
+results in a pod with a `nodeSelector` expression similar to what we did in the previous sections for the Spark jobs.
```
izmalk: Suggestion: let's move this result description before the command too.

izmalk (changes requested):
Nice work! It's very good. I've suggested a few wording and terminology improvements, but nothing critical. Mostly to make this guide sound more imperative and direct.
For example:
1. Update the page's title to sound like a guide.
2. Let's get rid of most "can" and "may" in this guide, especially when you introduce a command. Use the "to <achieve something> do <something>" pattern more instead.
3. Let's make sure we use the terminology here correctly in regards to capitalisation (I know it's a pain).
4. Consider dropping gerunds from headings.

docs/how-to/advanced-scheduling.md
Batalex: Thanks, I need to work on being more assertive in my writing for docs

docs/how-to/advanced-scheduling.md
Batalex: https://kubernetes.io/docs/reference/access-authn-authz/webhook/ has the two forms, I went with the capitalized one because I was refering to the general concept, not the API object (uncapitalized in the K8s docs).
What do you think is more relevant here? I'm happy to change it everywhere

docs/how-to/advanced-scheduling.md
Batalex: This model placeholder refers to the "Charmed Spark" charm control plane, where you would have deployed other charms. It could be the model for History Server, it could be the model for Kyuubi; it does not really matter.
I just thought it was not worth creating a brand new model for this one charm

docs/how-to/advanced-scheduling.md
Batalex: Thanks, done in 64a4b691e5a163f202b31542fc290a18194396e4.

docs/how-to/advanced-scheduling.md
Batalex: Done as part of 5d0c10d18181766c897401751f0e0b3064c9dd32

docs/how-to/advanced-scheduling.md
Batalex: I gave it a try in 5d0c10d18181766c897401751f0e0b3064c9dd32, is that the correct usage?

docs/how-to/advanced-scheduling.md
Batalex: and in 9fde56a6906ea87bac69222aae61ab51fa8ee1b5 as I missed your previous comment on the subject

docs/how-to/advanced-scheduling.md
Batalex: Should be done with 64a4b691e5a163f202b31542fc290a18194396e4 followed by 9fde56a6906ea87bac69222aae61ab51fa8ee1b5

docs/how-to/advanced-scheduling.md
izmalk: Looks great to me. Good job!

docs/how-to/advanced-scheduling.md
izmalk: I see now. I'm concerned that some users might have the same querstion arrise - I don't have a Charmed Apache Spark model; which one do I use?

Suggestion: Let's drop a model selector here. So that the command would just deploy the charm to a model, without specifying which one it needs to be. I think it's the closest to what we want, and on top of that - it simplifies the command.


```suggestion
juju deploy namespace-node-affinity --trust
```

docs/how-to/advanced-scheduling.md
izmalk: It looks like the page in their docs actually uses three different capitalisations of the word, which is kind of funny =D. To me, the lowercase version feels like the best fit, but it’s a small detail that depends on deep meaning and is totally up to you.
To me, it feels like we are talking about an instance of the concept here (because it has a state - "not configured to modify"), which makes me want to use lowercase.

izmalk (commented):
Looks great now. Just the model selector thing we need to resolve.

deusebio (approved):
Looks very good and thorough! Well done!

docs/how-to/advanced-scheduling.md
```diff
@@ -0,0 +1,254 @@
...
+        operator: In
+        values:
+        - arm64
+```
+
+```{note}
+Charmed Apache Spark provides multi-architecture rock images supporting amd64 and arm64.
+```
+
+### Defining a Pod template (alternative)
```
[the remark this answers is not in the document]
deusebio: That's fine. That's a bit of a broader discussion. Anyway, the presentation is now providing both option, and it is fine to not give particular nudges, also based on yes, possible conflicts. I'll resolve this now

izmalk (approved):
LGTM.

docs/how-to/advanced-scheduling.md
Batalex: Addressed in aed1e6b

docs/how-to/advanced-scheduling.md
Batalex: Addressed in 7a615705428ea2b5eb2328643534734a00f163e2

PR #205, opened 2026-03-24
[PRA-232] Pinning Juju 3.6.14 in CI

deusebio:
## Description

Current PR workarounds the current limitation of not being able to bootstrap controller previous of 3.6.18 with the latest snap. Therefore, after setting up the environment, we destroy controller and refresh the snap to a previous revisions. We then bootstrap a new controller

### DIFF
--- a/.github/workflows/ci-uat.yaml
+++ b/.github/workflows/ci-uat.yaml
@@ -37,6 +37,7 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v6
       - name: Setup tmate session
+        if: false
         uses: canonical/action-tmate@main
         with:
           detached: true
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-uat.yaml
```diff
@@ -36,6 +36,11 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v6
+      - name: Setup tmate session
```
welpaolo: Can we remove this ?

welpaolo (approved):
LGTM

.github/workflows/ci-uat.yaml
deusebio: yes, sure!

PR #208, opened 2026-04-02
[PRA-234] Terraform refactor following CC008

Batalex:
## Description

This PR is a big refactor of the bundles following CC008.

**Highlights**:
- Terraform modules are now under `/terraform`, + removed the previous ones. The resources previously present in the `releases/3.4` and `releases/3.5` folders now live under one `/resources` directory (the grafana dashboard was the same for 3.4 and 3.5, we'll see if we need to have a dedicated one for 4.0)
- Removed the yaml bundles. This means both the actual template files originally present under 3.4 and 3.5 but also the related python code and fixtures. We could streamline even more (no need for `BundleBackend` and the enum now that we only have TF). Let me know what you think.
- Migrated to Juju TF provider >1.0.0
- New reference docs automation using https://terraform-docs.io/ 

### How to navigate the new TF bundle
CC008 defines three types of TF bundles: charms, components and products:
```text
 terraform
 ├── charms
 │   ├── azure-storage-integrator
 │   ├── data-integrator
 │   ├── s3-integrator-v1
 │   └── zookeeper
 ├── components
 │   ├── observability
 │   ├── spark-3.4
 │   └── spark-3.5
 └── products
     ├── charmed-spark-3.4
     └── charmed-spark-3.5
```
The `charms/*` modules should be contributed to the upstream projects after some cleaning and more work on them.
I chose to have an `observability` component module instead of 4 separate charms module for convenience. Even if we do not maintain those charms, they more or less share a common domain, a release cycle and they don't have their own Juju model (hence, not a product module).

Right now we have a bit of duplication between the 3.4 and 3.5 components and products (but way less than before). We could factorize this in one module with a `flavor` or `major_minor` variable to parametrize the deployment, but I like explicitely having the track in the module URL for deployments refering to this repo, e.g. `source  = "git::https://github.com/canonical/spark-k8-bundle//terraform/products/charmed-spark-3.5?ref=mytag"`

### Testing and COS

The testing strategy needed to change a bit. Since a product takes care of its own model, testing multi-model deployment (such as spark + cos) requires having a single entry point defining the products as child modules. 
Therefore, we have a somewhat hidden TF module under `python/tests/integration/resources`, which is copied at test time in the temporary folder where we also copy our modules.
As a consequence, we can hardcode the storage sizes testing profile there.

COS-lite has a TF module, but we cannot really parametrize the channel/risk as of today. This does not seem to be an issue right now since we still get green CI, but we should ask the Obs team if they are ok with giving us a bit more leeway here so that we use a stable risk.
At the very least, this was a good opportunity to test that our solution properly integrates with the latest COS track 2.

### Documentation

I was unhappy for quite some time about our tooling to document the TF modules, so I tried something new in this PR. You can install the `terraform-docs` snap and enjoy the new `generate-docs` tox target, which takes care of updating the READMEs in all TF modules while keeping the manual documentation intact (only the part between the `<!-- BEGIN_TF_DOCS --> ... <!-- END_TF_DOCS -->` in the READMEs gets updated).


terraform/products/charmed-spark-3.4/outputs.tf
```diff
@@ -0,0 +1,53 @@
...
+output "metadata" {
+  description = "Metadata of the product deployment."
+  value = {
+    version     = "3.4"
+    deployed_at = timestamp()
+    updated_at  = timestamp()
+  }
+}
+
+output "model" {
```
deusebio: **todo** this should be models, and be an map. The current signature seems wrong. I understand here we only have one module, but I'm thinking that certain product modules may be more complex and have more than one module. So it is important to have an extensible signature that also cover those cases.

terraform/products/charmed-spark-3.4/outputs.tf
```diff
@@ -0,0 +1,53 @@
...
+    concat(
+      module.spark[*].offers
+    )...
+  )
+}
+
+output "metadata" {
+  description = "Metadata of the product deployment."
+  value = {
+    version     = "3.4"
```
deusebio: **suggestion** I would maybe concatanate this with also the tf versions of the module, such that this "version" becomes unique. Right now I would be concerned that this is way to slack, to allow to have reproducible versioning. 

I have not provided much guidance on this in the spec, and hence this is just a suggestion, but I would nudge towards having a more strict and precise versioning syntax

terraform/products/charmed-spark-3.4/outputs.tf
deusebio, line 1: **suggestion** In the outputs of the product modules I would possibly start to exercise also the `credentials` pattern, if this is easy enough. It feels that in Kyuubi we could expose a credential field with endpoins, username and password already, as result of the deployment.

terraform/products/charmed-spark-3.4/variables.tf
```diff
@@ -0,0 +1,279 @@
...
+  default     = {}
+}
+
+variable "integration_hub_revision" {
+  description = "Charm revision for spark-integration-hub-k8s"
+  type        = number
+  default     = null
+}
+
+variable "K8S_CLOUD" {
```
deusebio, line 108: **question** This (and the `K8S_CREDENTIAL` below) is very interesting. I'm wondering if this is an optional parameter or something we should require? If we require, I'm thinking that maybe the `juju_controller` parameter may be a possible place to bundle all the "cloud" input variables. 

But let's start first from the use-case to understand where this may go, and whether we should also update the spec and provide a bit more guidance on this

terraform/products/charmed-spark-3.4/variables.tf
deusebio, line 1: **discussion** Overall, we don't provide much guidance on the structure for the input variables, however when looking at this PR, I feel that we have a lot of variables flat in the root, while maybe using a bit of hierarcy may provide more logical segregation of the variables. For instance we could have an overall section "apps" where we provide the specification of the revisions/images. It feels a bit strange to provide input variables very granular, e.g. revisions, with very high level ones, e.g. K8S_CLOUD or `juju_controller`. 

Anyhow, this is not something where there is clear guidance, so in the sake of the PR, we can also keep it as-is, and then discuss and eventually move to a different settings in a dedicated ticket. In this thread, it would be anyway good to gather thoughts, and opinions, such that we can then streamline the separate discussion

terraform/products/charmed-spark-3.4/variables.tf
```diff
@@ -0,0 +1,279 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
```
deusebio, line 3: **todo** The only mandatory input that seems to be missing is `logging-config`

terraform/products/charmed-spark-3.4/variables.tf
```diff
@@ -0,0 +1,279 @@
+# Copyright 2024 Canonical Ltd.
```
deusebio: **todo** please fix the copyrights. I believe Rob pointed out last week. It is not a bit point, but it should take little (maybe even for copilot) to fix this

deusebio (commented):
This has been a massive refactoring. Well done Alex! Also getting the CI green and happy with such a large refactoring is definitely worth praising and noticing!

This is already quite a substantial step forward towards having better and more structured TF modules. So I would not be too nitpicking too much and block this PR for too long, since it already provides the base for a number of subsequent work, while still providing a sensible refactoring which aligns with CC008. However, I would possibly not merge it straight into main, and probably rather use a feature-branch and also land into that feature-branch subsequent improvements. Once the work is solid and finalized, we can then also merge it, but giving the heads up to a bunch of people first given that this refactoring changes substantially the API on how to interact with the TF module (and its dependency, e.g. using tf juju provider on 1.0.0 and onwards).

So, attached a have a few small and minor comments that should be easy to address before merging the current PR. Beside these, I have more substantial discussions/points, that however we could also handle as next steps, as they can be done in parallel or separately, for which the current PR represents a great base of work. 

The main points (to be addressed in self-contained and separate tickets/PR) I have are:

1. Given we are doing a substantial refactoring, why not also swapping to take the chance and swap `grafana-agent-k8s` with `opentelemetry-collector`?
2. I believe we are using `_endpoints` a little too much, and I would like to start using the framework for relations that can be either internal (using endpoints) or external (using offers). This is also to test this pattern, that I would be quite keen to use a speciement project (like spark) to test the pattern and its usability. Beside some of the services here (self-signed or postgresql) may indeed be in external models, when we use a centralized setup instead of a distributed one (where each application has its own postgresql and certificate provider)
3. I have read the PR comment, and I'm aware of your concerns about bundling 3.4 and 3.5 into the same product. On the product side, I believe you *may* have a point, and I don't have too strong opinions. For the user facing module, it may have sense to have two different products, but I would REALLY not have two modules for the components. This should not be as user-facing, and I'd rather have one single component (given how similar the two components are), and also given that these two versions will likely share the same lifecycle, at least from a tf module perspective (if we add a relation to one, we may well also add another relation to the other one as well, provided how similar they really are). On the component part, we could probably just swap between the two based on the `kyuubi` charm track, and we can JUST put some validation to make sure that the track here is either `3.4` or `3.5`. We could also put some validation for the track for `history-serfver` and `integation-hub` to make sure they are `3`. At the product level, we could start having two product modules, that however encode the logic to hardcode the right channel/track. I would probably also at the product level bundle the two with an input variables that automatically set the right track/channel (for lower maintainability burden, given they would be very similar), but I'm less opinionated on this. 
4. I would remove the `locals.tf` where we pin revisions and images. I understand it is the current setting, but it is a bit of a mess when dealing with different architectures, etc. You have done the right thing to not overload the current PR with this, but I believe it would be nice to move away from this pattern in a subsequent PR. While doing this removal, we could also introduce the input variable `risk` a bit like we have done in KF, but let's discuss about this whether this is a pattern we like  and we are keen on (spoiler alert: I kinda do, but let's discuss)
5. Adding support for different architectures: kinda related to the previous point (since the pinning of revisions kinda prevents this today), I would also like to allow to support deployment on different architectures, possibly using some high-level variables. This should definitely be allowed at `product` level (where we can have on high-level input variable to allow to swap from one to another), I'm less opinionated on the `component` level, where this can be set using the various `charms.constraint`. One pattern we could provide is to both allow a high-level `architecture` value (if you specify this, all the possible charms that support that architecture will be setting the constraint) while still allow some more granual tuning, something like:
```
constraints = var.history_server.constraints != null ? var.history_server.constraints else "arch={var.architecture}"
```
Just a thought, but given we know support arm, we should definitely have something like this to provide an easy way to test this (possibly using solqa and mixed architecture there) given testing arm on runner is not possible untill all charms in the Kyuubi bundle will support arm.

terraform/products/charmed-spark-3.4/outputs.tf
Batalex: I don't get your second point, as it should follow CC008's structure:

> Map of the key of the model and the components deployed in the model, e.g. 
>```
>type = {
>  model_uuid = …
>  components = {
>    charm1 = …
>    charm2 = …
>  }
>}
>```

Did you have in mind something like this instead?
```
{
<model_uuid1> = {
    components = {
        charm1 = ...
        charm2 = ...
},
<model_uuid2> = {
    components = {...}
}
};
```

terraform/products/charmed-spark-3.4/variables.tf
Batalex: Updated copyrights in the product modules to 2026 in f5a0942e0efd74bb9ddb1a00966308e5d2619990

terraform/products/charmed-spark-3.4/outputs.tf
Batalex: Fixed output name in 701be455602fd7a0d2592b5e5344a54a6edad1b1

terraform/products/charmed-spark-3.4/outputs.tf
Batalex: I can see how we'd expose the username (hardcoded) but how do we get the password when it's not user-provided, or the jdbc endpoint? They are only available through an integration

.github/actions/run-test/action.yaml
```diff
@@ -23,11 +23,6 @@ inputs:
     default: 3.4.4
     required: false
     type: string
-  bundle-backend:
```
theoctober19th, line 26: praise: i was waiting so long for us to remove the YAML backend. feels so good to see we finally remove this.

terraform/charms/azure-storage-integrator/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+  description = "The operating system on which to deploy"
+  type        = string
+  default     = "ubuntu@22.04"
+  nullable    = false
+}
+
+variable "channel" {
+  description = "Channel of the charm."
+  type        = string
+  default     = "latest/edge" # TODO: Update to stable once we have the new release
```
theoctober19th: suggestion: Use `1/edge` here, since `latest` track will no longer receive updates.

theoctober19th (commented):
I had a comprehensive look of the PR once, and thus with the first impression, really nice work done. I specially liked the idea to remove the YAML bundle (which is deprecated)

terraform/charms/azure-storage-integrator/variables.tf
Batalex: Bumped in 5d4cd55cfc6723060fb48fa2b2e4c3ac382838ff

terraform/products/charmed-spark-3.4/variables.tf
Batalex: Added logging configuration to the model in fd587457934eb54ed7f0f6561075b1e1d0b86a12

terraform/products/charmed-spark-3.4/variables.tf
Batalex: I removed those for now, since we target a feature branch we can think about this a bit more

terraform/products/charmed-spark-3.4/outputs.tf
Batalex: Fixed output format in dadf44a37f1aede5fd95e912654d63cb8ac07cf0

terraform/products/charmed-spark-3.4/outputs.tf
Batalex: Added a simple placeholder in 07a893c60386f0c3acd22a6edb15e860f3d8705a, we can discuss in a subsequent PR how we will actually fill that placeholder

terraform/products/charmed-spark-3.4/outputs.tf
deusebio: Yeah, good point. There may be ways (e.g. using tf actions or some nil resources), but it is not for this PR to figure it out. Let's just defer this to a dedicated ticket. It was also just a suggestion, not really blocking. Resolving the comment now

terraform/products/charmed-spark-3.4/variables.tf
deusebio: Yes, sounds good!

theoctober19th (approved):
Great work Alex, let's merge to feature branch and then improve it in coming iterations.

python/tests/integration/resources/main.tf
```diff
@@ -0,0 +1,125 @@
...
+  name       = var.spark_model_name
+  credential = var.K8S_CREDENTIAL
+  cloud {
+    name = var.K8S_CLOUD
+  }
+}
+
+module "spark" {
+  depends_on = [juju_model.spark, module.cos]
+  source     = "./products/charmed-spark-<spark_flavor>" # filled by test fixture
```
welpaolo, line 99: Should be more appropriate to call it spark versions ?
```suggestion
  source     = "./products/charmed-spark-<spark_version>" # filled by test fixture
```

terraform/products/charmed-spark-3.4/providers.tf
```diff
@@ -0,0 +1,10 @@
+# Copyright 2026 Canonical Ltd.
```
welpaolo, line 1: Do we need this file ?

welpaolo (approved):
Great work Alex :) 
I left a couple of minor and non blocking comments.

terraform/products/charmed-spark-3.4/providers.tf
Batalex: The juju provider block is part of the spec, but it causes an error when trying to import the module from another module (like this is the case for deployment modules). I wrote a comment on the spec with this finding and commented the code here while we settle on the matter.

python/tests/integration/resources/main.tf
Batalex: Yes, it's a better name; I'll change it

Batalex:
Thanks for the reviews everyone, I'll merge this on the feature branch so that subsequent PRs (and reviews) can be smaller, more focused and as nitpicky as we want them to

PR #215, opened 2026-04-17
CC008 Spec alignment

Batalex:
This PR addresses the following, previously missing from #208:
- `deployed_at` product output 
- swap-able external relations

Other changes:
- We now use COS lite components from 2/stable instead of the dev/ track
- We introduced tflint to lint the terraform modules. This PR removes some variables we might re-add later, that's fine.
- The tox target for terraform-docs was renamed to `tf-docs`, shorter to type and more explicit

### `deployed_at`

On the first apply, `terraform output` gives the same timestamp for both this field and `updated_at`. However, the `terraform_data` resource will make sure that the value is tracked in the terraform state and will never be updated. Therefore, subsequent `terraform apply` will leave this attribute untouched.

### Component external relations

Relations with postgresql (x2), certificate provider, zookeeper and data-integrator are now following the framework defined in CC008 and should be easily swap-able for cross-model offers. However, this is not tested in CI since this would make our tests even more complex.
The result is a bit verbose, as I added some validation rules for each integration.
We also had to introduce a new component variable `object_storage_interface` since we cannot rely anymore exclusively on the object storage endpoint string value (undefined for a cross-model offer) to use the proper interface on the other end of the relation.


.github/workflows/ci-checks.yaml
```diff
@@ -31,10 +31,17 @@ jobs:
...
+      - uses: terraform-linters/setup-tflint@v6
+        name: Setup TFLint
+
       - id: code-checks
         name: Make sure the code follows conventions and tests correctly
         run: |
-          cd python && tox -e lint
+          cd python
+          tox -e lint
+          tox -e tf-lint
```
deusebio: **suggestion** why not having only ONE lint environment and doing both linting for python and tf? I don't have strong opinions, but I also don't see the use-case that we want to NOT run both linting together

terraform/charms/azure-storage-integrator/variables.tf
```diff
@@ -8,13 +8,6 @@ variable "app_name" {
   nullable    = false
 }
 
-variable "base" {
```
deusebio, line 11: **question** are we removing this because azure storage integrator is ONLY provided on 22.04? If so, it seems sane. Eventually when more than one base is supported, we could readd this. It should nevertheless not break backwards compatibility

terraform/products/charmed-spark-3.4/main.tf
```diff
@@ -207,24 +203,25 @@ module "spark" {
...
 
-  tls_endpoint = {
+  certificates = {
+    kind     = "endpoint"
     name     = module.ssc.app_name
     endpoint = module.ssc.provides.certificates
   }
-
-  metastore_endpoint = {
+  data_integrator = merge({ kind = "endpoint" }, module.data_integrator.requires.kyuubi)
```
deusebio, line 211: **thought** it is fine since it is compliant to the spec, but I'm thinking that we could possibly discuss with the TF group to change the signature of the `requires` at the source. It is a bit ugly to always have to do this reformatting.

terraform/products/charmed-spark-3.5/outputs.tf
```diff
@@ -15,7 +15,7 @@ output "metadata" {
   description = "Metadata of the product deployment."
   value = {
     version     = "3.5-tf_0.0.1"
-    deployed_at = timestamp()
+    deployed_at = terraform_data.deployed_at.output
```
deusebio, line 18: **praise** very nice! Well done!

terraform/products/charmed-spark-3.5/variables.tf
```diff
@@ -57,18 +52,6 @@ variable "cos_offers" {
   default = null
 }
 
-variable "data_integrator_revision" {
-  description = "Charm revision for data-integrator"
-  type        = number
-  default     = null
-}
-
-variable "grafana_agent_revision" {
```
deusebio, line 66: **question,issue** why have we removed these revisions? I don't think we should remove them. I mean the ability to specify revisions does not only apply to Spark components, but to all components, see in the spec:

> Allow specification of the charm revision for every single charm. The default value for the charm revisions should be null, such that the latest revision on a given channel is used
> Allow specification of the images to be used for every single charm. The default value can be null, therefore following back on using the resource specified by the charm.

This way the product modules can also easily be used to test new revisions of Postgresql, data-integrator, etc. Also from a consumer of product perspective, this is a bundle. I don't think we should treat "our" charm in any special manner

deusebio (changes requested):
The PR contains a number of sensible changes, that also trigger a few thoughts on my part.

I only have one major comment about the fact that it seems that we are not able to specify the revisions for a few charms that do not belong to the product. Also according to the spec, the product module should allow this for every charms that it bundles.

terraform/charms/azure-storage-integrator/variables.tf
Batalex: No, I only removed those because they were unused variables so tflint was complaining. I did not think much of it since those modules are bound to be removed from here in the medium term

.github/workflows/ci-checks.yaml
Batalex: Yeah, I liked the idea of having `tox -e tf-<whatever>` for everything affecting the modules but let's remove duplicates. Module formatting was handled in the same target as python files', and that was working fine for us

terraform/products/charmed-spark-3.5/variables.tf
Batalex: Fair enough, this was just done to get the linter happy until we wrap up the tfvars file approach and avoid PR scope creep. I intended to add those again once I figured out exactly how and where we should generate this file. I'll try to address this in this PR

terraform/products/charmed-spark-3.4/main.tf
Batalex: Honestly I don't mind, still better than having to do the ternaries because of `count` on optional modules IMO. I added a comment in the spec so that we can discuss it with the working group

.github/workflows/ci-checks.yaml
Batalex: Merged the two targets in 79fcd5ac5f01b63a0ee

terraform/charms/azure-storage-integrator/variables.tf
Batalex: Properly re-added the variables in 16f9f5f127c64f90. It was just faster at the time to remove everything that made the linter unhappy, but when it is time to contribute those modules back to the projects, we might as well have something solid from the get-go

terraform/products/charmed-spark-3.4/main.tf
Batalex: At least we removed a fair amount of ternaries in 8f6f9234e5999839263b07e00a68d5043c980e30

terraform/products/charmed-spark-3.5/variables.tf
Batalex: Re-added the variables in f8c20a9a384ccdb3e15bc4969c4b14e926e4a5fc, this time properly wired up to the final tf resources

terraform/products/charmed-spark-3.4/main.tf
deusebio: Nice!

deusebio (approved):
Looks really good! Great job Alex!

terraform/components/observability/README.md
```diff
@@ -59,7 +59,7 @@ No modules.
...
+| <a name="input_cos_configuration"></a> [cos\_configuration](#input\_cos\_configuration) | n/a | <pre>object({<br/>    app_name = optional(string, "cos-configuration")<br/>    base     = optional(string, "ubuntu@22.04")<br/>    channel  = optional(string, "1/stable")<br/>    config = optional(map(string), {<br/>      git_branch              = "main"<br/>      git_depth               = 1<br/>      git_repo                = "https://github.com/canonical/spark-k8s-bundle"<br/>      grafana_dashboards_path = "resources/grafana/"<br/>    })<br/>    constraints = optional(string, "arch=amd64")<br/>    resources   = optional(map(string), {})<br/>    revision    = optional(number, 65) # TODO<br/>    units       = optional(number, 1)<br/>  })</pre> | `{}` | no |
```
welpaolo, line 62: It this `TODO` left on purpose here as well ?

welpaolo (approved):
LGTM

terraform/components/observability/README.md
Batalex: It's automatically picked up by `terraform-docs`, it's quite odd indeed. Anyhow, it will go away once we settle on the design decision of where the revisions must be pinned and drop the TODO comment

PR #219, opened 2026-04-21
[PRA-231] Update bundle with updated candidate versions

deusebio:
## Description

We align with latest stable changes as well as latest changes that have landed in the `edge` channels for the Charmed Spark charms.

## Checklist

- [ ] I have added or updated any relevant documentation.

### DIFF
--- a/releases/3.4/terraform/versions.tf
+++ b/releases/3.4/terraform/versions.tf
@@ -20,7 +20,7 @@ locals {
     zookeeper         = 78
     data_integrator   = 362
     s3                = 330
-    azure_storage     = 270
+    azure_storage     = 274
     grafana_agent     = 121
     # TODO: bump the revision to 1/stable when both of the following issue gets fixed:
     # https://github.com/canonical/cos-configuration-k8s-operator/issues/128 
--- a/releases/3.4/terraform/versions.tf
+++ b/releases/3.4/terraform/versions.tf
@@ -12,20 +12,17 @@ terraform {
 
 locals {
   revisions = {
-    history_server  = 96
-    integration_hub = 119
-    kyuubi          = 149
-    kyuubi_users    = 495
-    metastore       = 495
-    zookeeper       = 78
-    data_integrator = 362
-    s3              = 330
-    azure_storage   = 270
-    grafana_agent   = 121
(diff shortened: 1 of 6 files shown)


releases/3.4/terraform/versions.tf
```diff
@@ -12,16 +12,16 @@ terraform {
...
-    grafana_agent   = 121
+    history_server    = 98
+    integration_hub   = 123
+    kyuubi            = 162
+    kyuubi_users      = 774
+    metastore         = 774
+    zookeeper         = 78
+    data_integrator   = 362
+    s3                = 330
+    azure_storage     = 270
```
theoctober19th: suggestion: use revision `274` from `1/candidate`.

releases/3.5/terraform/versions.tf
```diff
@@ -12,16 +12,16 @@ terraform {
...
-    grafana_agent   = 121
+    history_server    = 98
+    integration_hub   = 123
+    kyuubi            = 160
+    kyuubi_users      = 774
+    metastore         = 774
+    zookeeper         = 78
+    data_integrator   = 362
+    s3                = 330
+    azure_storage     = 270
```
theoctober19th: suggestion: use revision 274 from 1/candidate.

theoctober19th (commented):
The versions and revisions look good to me. I have a suggestion that perhaps with this PR we can also update `azure-storage-integrator` to use `1/edge` track instead of `latest/edge`, since in the future that track will receive no updates.

deusebio:
@theoctober19th I initially avoided to update this given it was not YET in stable, but you are right, better to start using the `1/` track. It is nevertheless better than `latest/edge`

Batalex (approved):

theoctober19th (approved):

releases/3.5/terraform/versions.tf
deusebio: Done!

releases/3.4/terraform/versions.tf
deusebio: Done!

welpaolo (approved):
LGTM

PR #221, opened 2026-04-23
[PRA-11] Automated tutorial testing

izmalk:
## Description

Created an automated solution for the tutorial end-to-end tests using Spread + Multipass.

1. The `extract_commands.py` script parses the existing tutorial content and extracts necessary commands in correct order.
2. The tutorial files are modified with special comment-blocks, invisible to users, that adjust the flow for automated run when needed.
3. Spread tasks are also generated from tutorial files.
4. Spread.yaml to run it either locally (tox -e tutorial) or via CI/CD (this PR does not include CI/CD).
5. The tests are run in a Multipass VM.

Estimated time from start to finish ~1-1.5 hour.

The output finishes like this:

```text
+ multipass delete --purge spark-tutorial-vm

-----
2026-04-25 16:03:54 Allocation results of multipass:ubuntu-24.04-64: map[string]string{}
2026-04-25 16:03:54 Worker terminated.
2026-04-25 16:03:54 Successful tasks: 7
2026-04-25 16:03:54 Aborted tasks: 0
  tutorial: OK (4368.68=setup[0.02]+cmd[0.02,4368.64] seconds)
  congratulations :) (4368.74 seconds)
```

I've added TESTING.md with the description of the solution and updated the README.

## Architecture explanation

Our goal in this PR is to add end-to-end testing for the tutorial and verify that users can complete it successfully.
For example, I've already noticed and fixed a small problem with S3 credentials propagation via the Integration hub in the tutorial.

The tests intentionally rely on the exact code blocks that we ask users to run. For that reason, this implementation uses shell-based logic rather than Python/Jubilant: it keeps the test flow as close as possible to the actual user experience.

Additionally, we use Spread for testing infrastructure as that is the recommended option and it works well with shell-based logic.

One trade-off is that the shell implementation of wait_idle() is not especially elegant and may be harder to maintain over time. I have already started prototyping alternatives based on Jubilant and juju wait-for. However, initial testing showed that the shell script is currently the most reliable and consistent option.

For this PR, I would like to merge the reliable shell-based implementation as a working baseline first. I will continue exploring the more maintainable alternatives separately, and once we decide which implementation serves us best I will refactor the testing automation outside the scope of this PR.

## Checklist


python/tests/tutorial/resources/count_ubuntu.py
```diff
@@ -0,0 +1,28 @@
+from operator import add
```
welpaolo, line 1: you can add copyright at the beginning of each python file:
```
#!/usr/bin/env python3
# Copyright 2026 Canonical Ltd.
# See LICENSE file for licensing details.
```

docs/tutorial/2-distributed-data-processing.md
```diff
@@ -132,6 +137,20 @@ Let's download a sample dataset, store it in an S3 object storage, and do some d
 
 #### Prepare a dataset
 
+<!-- test:run
+# Create a mock twitter.csv for automated testing (avoids 500MB Kaggle download)
+mkdir -p twcs
+cat > twcs/twcs.csv << 'MOCK_CSV_EOF'
```
deusebio, line 143: **suggestion** we could possibly have the csv file in the `tests/tutorial/resources/twitter.csv` and here just copy that file without creating a file with the command line (which also does not match with the one in the tutorial).

docs/tutorial/2-distributed-data-processing.md
```diff
@@ -216,39 +235,17 @@ Charmed Apache Spark comes with a command that can be used to submit Spark jobs
...
-    rdd.map(lambda row: row["text"])  # Extract the "text" column
-        .map(contains_ubuntu)  # Apply the contains_ubuntu function
-        .reduce(add)  # Sum all the 1 and 0 to get the total number of matches
-)
-
-# Print the result
-print(f"Number of tweets containing Ubuntu: {count}")
-
-spark.stop()
+```{literalinclude} ../../python/tests/tutorial/resources/count_ubuntu.py
```
deusebio, line 238: **praise** very nice touch to use one single source of truth! Well done!

python/tests/tutorial/extract_commands.py
deusebio, line 1: **suggestion** it would make sense to factor this out in its own package that can be simply installed with `pip install ...` instead of copy and pasting this into the various repositories (such that you can also write unittests/integration tests for this tool)

deusebio (approved):
Looks very good! Having a tested tutorial is extremely sane and cool! I don't have many concern with the approach of using shell instead of jubilant, and I'm not sure why one should be preferred over the other. I believe Jubilant may be preferred when we write the tests ourselves, to use python which is generally more maintainable than bash. But here, the backend (either bash or python) is something abstracted away from the user that is done by the compilation:

tutorial => compiled code => test execution

Whether the compiled code and the test execution are based on bash or python, this is an implementation detail, with no reason why one should be preferred over the other. 

Maybe the only comment here could rather be that the python scripts that do the translation and "compilation" of the documentation into an executable that can be tested could be factored out into its own library/tool, such that the codebase does not need to be copied and pasted across the various repositories. When doing this refactoring, one could also possibly create an API to translate the docs into a testable file, and then provide also the tooling to run the file. In this way, one could also possibly change the underlying engine from bash to python/jubilant, with no impact on the downstream projects. But still, I'm not sure what the value of swapping from shell to Jubilant can be.

python/tests/tutorial/extract_commands.py
izmalk: Yes, I was thinking about that.
Maybe do it as a CI/CD action too.
But we'll have to experiment with this next cycle, when I'll release it as a standalone thing.

docs/tutorial/2-distributed-data-processing.md
izmalk: > which also does not match with the one in the tutorial

Sorry, I'm a bit confused, but we have such file in the tutorial - it's the one being extracted from the downloaded archive. So by using the same filename I try to preserve consistency and make it traceable to the real tutorial. Even the S3 command is practically the same:

```bash
aws s3 cp ./twcs/twcs.csv s3://spark-tutorial/twitter.csv --checksum-algorithm SHA256
```

docs/tutorial/2-distributed-data-processing.md
```diff
@@ -305,7 +302,23 @@ The executor pods are deleted automatically.
...
-```bash
+<!-- test:wait --seconds 60 -->
+
+<!-- test:run
+# Wait for the driver pod to reach Completed state
+for i in $(seq 1 30); do
+  pod_name=$(kubectl get pods -n spark | grep "count-ubuntu-.*-driver" | tail -n 1 | cut -d' ' -f1)
+  if [ -n "$pod_name" ]; then
+    phase=$(kubectl get pod "$pod_name" -n spark -o jsonpath='{.status.phase}' 2>/dev/null)
+    if [ "$phase" = "Succeeded" ] || [ "$phase" = "Failed" ]; then
```
Batalex, line 313: **question**: Should we only target job success?

**suggestion**: `kubectl wait` might be simpler:

```shell
kubectl wait --for=condition=Complete --timeout=150s pod/${pod_name} -n spark
```

docs/tutorial/3-data-stream-processing.md
```diff
@@ -66,9 +82,9 @@ Once all the charms have been deployed and integrated (you may need to wait some
 Model            Controller      Cloud/Region        Version  SLA          Timestamp
 spark-streaming  spark-tutorial  microk8s/localhost  3.1.7    unsupported  10:13:55Z
 
-App            Version  Status  Scale  Charm          Channel  Rev  Address         Exposed  Message
-kafka-k8s               active      1  kafka-k8s      3/edge    47  10.152.183.242  no       
-zookeeper-k8s           active      1  zookeeper-k8s  3/edge    42  10.152.183.87   no       
+App            Version  Status  Scale  Charm          Channel   Rev  Address         Exposed  Message
+kafka-k8s               active      1  kafka-k8s      3/stable   47  10.152.183.242  no       
+zookeeper-k8s           active      1  zookeeper-k8s  3/stable   42  10.152.183.87   no
```
Batalex: **note**: Not sure how relevant this is to the present PR: ZK 3/stable amd64 is currently rev 78, Kafka is 82. Moreover, the latest Juju stable cannot bootstrap a 3.1.7 controller.
We can put a note to update this kind of stuff in a subsequent PR, I still have in mind that we need to update the integration-hub monitored-service-accounts configuration option now that we released to stable

README.md
```diff
@@ -43,6 +43,24 @@ For information on security features and the use of cryptography, see the [Secur
...
 Security issues in the Charmed Apache Spark can be reported through [LaunchPad](https://wiki.ubuntu.com/DebuggingSecurity). Please do not file GitHub issues about security issues.
 
+## Tutorial tests
+
+The tutorial (`docs/tutorial/`) is tested end-to-end using [Spread](https://github.com/canonical/spread) inside a Multipass VM. Shell commands are extracted directly from the Markdown sources, so the tutorial itself is the test.
+
+Run the full tutorial suite (extract + run Spread tests):
+
+```bash
+tox -e tutorial
```
Batalex, line 53: **praise**: this is so cooooool

Batalex (approved):
This is so cool, I love it so much

welpaolo (approved):
Great work @izmalk :) I run it locally and it worked perfectly! Really great milestone for our documentation!

docs/tutorial/2-distributed-data-processing.md
izmalk: > question: Should we only target job success?

This loop's sole purpose is to wait for the pod to reach a terminal state, not to assert success. This way we can exit the loop faster than timeout.

> suggestion: kubectl wait might be simpler:

That's a great idea, thanks!

Unfortunately it's not as straightforward implementation-wise, for example, we would still need to get the pod's name, and K8s pods terminal condition is... tricky.
The best idea I have for now is to keep the cycle with the `pod_name` retrieval and kubectl wait on top of it, making it two waits instead of just one:

```text
# Wait for the driver pod to appear
for i in $(seq 1 30); do
  pod_name=$(kubectl get pods -n spark | grep "count-ubuntu-.*-driver" | tail -n 1 | cut -d' ' -f1)
  [ -n "$pod_name" ] && break
  sleep 5
done

# Wait for the driver pod to succeed
kubectl wait --for=jsonpath='{.status.phase}'=Succeeded --timeout=150s pod/"${pod_name}" -n spark
```

Let's discuss that further to design it better and implement in next iteration.

python/tests/tutorial/resources/count_ubuntu.py
izmalk: Fixed for the extractor script in https://github.com/canonical/spark-k8s-bundle/pull/221/commits/9fd25b0e72cdd3517723993035677d831c35b87f

As for the two examples here, it's tricky. As we don't want the copyrights to show up ion our docs taking space. I hope it's OK if we skip the copyright for such simple examples.

docs/tutorial/3-data-stream-processing.md
izmalk: Great catch. I've used test's output to update all output now in https://github.com/canonical/spark-k8s-bundle/pull/221/commits/315aae94c0c4f9c3ee77c550a021bbeaa98f4189

PR #223, opened 2026-05-05
Terraform risk input; remove version pinning from the modules

Batalex:
This PR addresses the following issues we've discussed:
- We no longer have revisions and images pinned in `locals.tf`. Everything at test time is provided by a tfvars-like mechanism. This will make it easier in the future to manage a greater combination matrix (3 tracks on 2 archs).
- Added a global `risk` for the spark component module. This led to a change, the various charms inside no longer expose a `channel` input but a `track` one, since we interpolate track/risk to create the channel. Please let me know if this is in line with what you had in mind

We mentioned having some kind of higher-level variable in the products module, like `apps`, because it feel a bit weird to have deployment-like variables mixed with very specific ones. I don't mind the situation as-is, because this way we can provide a proper description for each item. With a big `apps` mapping, we would only have a description for it, not for the nested parts. Wdyt?


deusebio (commented):
The changes are really going in a great direction, with very nice patterns and high-level UX. 

I have a few points to be discussed next week:

1. I would have added the `risk` variable only at `product` level, and not at component level. This in order to keep the signature of components more consistent with the charm modules. Interestingly, also observability team was suggeting to expose track and risk as variables instead of channel in charm modules. In general I don't have too strong opinions, but I would not change things unless requires at charm and component level. So basically, when you reference the spark component in the product module you could use string interpolations, e.g. 
```
module "spark" {
  ...
  source     = "../../components/spark"
  risk       = var.spark_risk
  ...

  history_server = {
    ...
    channel = "3/${var.spark_risk}"
  }
```

2. I know that I was the first introducing the YAML to provide revisions and images in the tests for the automated risk promotion, but while reviewing this PR I realized that we should probably use `tfvars.json` formats for the various tfvars, such that these could also be consumed natively by TF using `terraform apply -var-file 3.4_amd64.json`. As it stands right now, these files can only be consumed by the tests

3. It seems that we actually never do a tests with the `default` values. I'm thinking that it would probably be also sane to check if a deployment with no input variables works

Batalex:
I agree on 1. and 3.
About 2.:


>     2. I know that I was the first introducing the YAML to provide revisions and images in the tests for the automated risk promotion, but while reviewing this PR I realized that we should probably use `tfvars.json` formats for the various tfvars, such that these could also be consumed natively by TF using `terraform apply -var-file 3.4_amd64.json`. As it stands right now, these files can only be consumed by the tests

I actually started with a json file but went back to a yaml approach in 9239189. The main reason is that the json spec does not allow comments while it is not an issue with yaml. Do you still want to proceed, knowing that those comments allow us to document the rationale behind pinning a specific revision? Where else could we document that, and how doable would it be to maintain the revisions in one place and the explanations in another?

Batalex:
This new version for review:
- converts the pin files from yaml to hcl, so that they can be used by the terraform CLI (though their content is only compatible with the test entrypoint `tests/integration/resources/main.tf`)
- Adds a new test (and the needed machinery) to run a simple integration test using unpinned revisions/OCI resources in CI

deusebio (approved):
Looks very good! Well done @Batalex !

PR #224, opened 2026-05-15
[PRA-285] Add support for Spark 4 in Spark K8s Bundle

theoctober19th:
TLDR;
1. Add support for Spark 4 (starting from TF CC08 work)
2. Upgrade s3-integrator to track 2/ for all products (3.4, 3.5) and use track 2/ for the new 4.0 product as well.

Full description of changes:

This pull request introduces several significant improvements to the integration testing infrastructure and deployment configuration for Spark, particularly adding support for Spark 4.0, updating S3 integration, and improving credential handling. The changes span CI workflows, Terraform modules, and integration test fixtures.

**Key changes include:**

### Spark 4.0 Support and Test Matrix Updates

* Added Spark 4.0.2 to the supported versions in both full and minimal CI test matrices, and included `integration-basic` tests for Spark 4.0.2. Also, introduced a new tfvars file for Spark 4.0 with appropriate component revisions and images. [[1]](diffhunk://#diff-538a9b825fc59ef4f828a2e910c32ad1571e33127f97a14ebd296cf8968e413bR31-R35) [[2]](diffhunk://#diff-3be9733b31e57301ce7745c8e3e970737bba9d7bc1dd4eef5b921e6dfdecf10dL32-R32) [[3]](diffhunk://#diff-3be9733b31e57301ce7745c8e3e970737bba9d7bc1dd4eef5b921e6dfdecf10dR60-R67) [[4]](diffhunk://#diff-36393b6b4c9e8b7a0476a880e56b02217c71b1def2484bbd7ab28dbd5ef4d957R1-R20)
* Updated `specs.yaml` to define the correct channels for Spark 4.0 components.

### S3 Integration and Credential Management

[...]

PR #225, opened 2026-05-21
[TODO: Merge with History] Terraform refactor following CC008 and Spark 4 changes

Batalex:
This is a big piece; most of it has already been reviewed though.

Highlights: 
- New terraform modules following CC008
    - Charm modules to be upstreamed 
    - Two component modules: Spark and Observability. We have a WIP PR to migrate from grafana-agent to otelcol, but unfortunately it seems like we lose some labels in the loki log integration. To be resumed later
    - Two (for now) product modules: 3.4 and 3.5. Bikalpa is working on 4.0 using this new structure
- New integration test running without pinning the revisions, using the latest and greatest for all charmed-spark charms and oci resources
- New risk input that should facilitate testing before promotion


deusebio (approved):
I have already reviewed these changes in the various PRs on the feature branch. Looks good to me. 

It is probably worth also getting Dario's second look though, since this is effectively the complete feature branch

deusebio:
@dariofaccin

theoctober19th (approved):
Super awesome work! I have fixed a few typos, a few missing docs here and there via #224

theoctober19th:
Hi @Batalex, I see that once we change the TF structure, I think we'd also need a significant change in our docs on how to deploy our bundle using Terraform. How do we plan to tackle that task, should we perhaps create a ticket and work on it in the near future? cc: @deusebio

theoctober19th:
One other issue I noticed is that, while `tls-client-private-key` is an optional config option for `kyuubi-k8s` (and when this is not given, Kyuubi will auto generate its own private key and use it for TLS certificates), the bundle makes it sort of necessary.

This is because the bundle now uses the same secret for `system-users` and `tls-client-private-key`, and thus if in case the TLS key is not provided via tfvars, the config option `tls-client-private-key` is still supplied (but it will instead only have system users inside that secret) -- resulting in Kyuubi to complain "invalid tls key provided".

image

I'll create an issue to track this.

welpaolo (approved):

deusebio (approved):
Let's goooooo! I have already added the rule that should give you permission to push to `main`, such that you can rebase remotely

Issue #236, opened 2026-06-17
The `tls-client-private-key` is an optional config option for `kyuubi-k8s` but the bundle makes it necessary


theoctober19th:
Commented on #225:

One other issue I noticed is that, while `tls-client-private-key` is an optional config option for `kyuubi-k8s` (and when this is not given, Kyuubi will auto generate its own private key and use it for TLS certificates), the bundle makes it sort of necessary.

This is because the bundle now uses the same secret for `system-users` and `tls-client-private-key`, and thus if in case the TLS key is not provided via tfvars, the config option `tls-client-private-key` is still supplied (but it will instead only have system users inside that secret) -- resulting in Kyuubi to complain "invalid tls key provided".

<img width="1852" height="841" alt="image" src="https://github.com/user-attachments/assets/86d1219c-4b0c-4ef7-a46c-b42681ea4287" />

I'll create an issue to track this.

_Originally posted by @theoctober19th in https://github.com/canonical/spark-k8s-bundle/issues/225#issuecomment-4725883943_

PR #237, opened 2026-06-17
[PRA-309] Update docs for s3-integrator 2/stable and multi-track awareness

theoctober19th:
This PR aims to update the places where tags and tracks are specified to warn users to use correct tags and tracks for the version of Apache Spark they would like to use.

In the meanwhile, I have also done refactoring of the how-to-guide to deploy Charmed Spark solution.

Also update the docs to remove reference to `s3-integrator` from `1/stable` and use `2/stable` instead.


docs/how-to/deploy/kyuubi.md
```diff
@@ -71,22 +72,30 @@ Please see the following sections for guidance on how to configure each.
 
 ### S3
 
-For S3-compliant storage, deploy [S3 integrator](https://charmhub.io/s3-integrator):
+For S3-compliant storage, deploy [S3 integrator](https://charmhub.io/s3-integrator) from channel `2/stable`:
 
 ```bash
-juju deploy s3-integrator
+juju deploy s3-integrator --channel 2/stable
```
Batalex, line 78: **thought**: Any idea when we'll be able to set track 2 as the default one?

docs/how-to/deploy/kyuubi.md
```diff
@@ -71,22 +72,30 @@ Please see the following sections for guidance on how to configure each.
...
+```bash
+juju add-secret s3-creds access-key=<ACCESS-KEY> secret-key=<SECRET-KEY>
+# secret:jem6a4josup6rui0g2q0  <-- make note of this secret URI
+
+juju grant-secret s3-creds s3-integrator
 ```
 
 Configure S3 integrator to use your S3 object storage by setting up endpoint address, bucket, path to folder, and credentials:
 
 ```text
```
Batalex: **nitpick**: No change required. We could be more consistent with the code blocks, here we are using `text`, but `bash` in the ones just before and after. 
I also believe that Vladimir is more keen on `shell`

docs/how-to/deploy/kyuubi.md
```diff
@@ -96,23 +105,27 @@ After that, `kyuubi-k8s` app becomes blocked with the `Missing authentication da
 
 ### Azure DataLake
 
-For Azure DataLake storage, deploy [Azure storage integrator](https://charmhub.io/azure-storage-integrator):
+For Azure DataLake storage, deploy [Azure storage integrator](https://charmhub.io/azure-storage-integrator) from channel `1/stable`:
 
 ```bash
-juju deploy azure-storage-integrator
+juju deploy azure-storage-integrator --channel 1/stable
```
Batalex, line 119: **thought**: Same as for s3, any idea when we will be able to set 1 as the default track? + `latest` does not have a stable release so it would be even more useful here

docs/how-to/deploy/spark.md
```diff
@@ -28,223 +28,208 @@ an S3-compatible object storage on MicroK8s (MinIO), EKS (AWS S3), or Azure obje
...
-Juju bundles are provided in the form of Jinja2 templates, for the following distribution:
+module "spark" {
+  source     = "git::https://github.com/canonical/spark-k8s-bundle//terraform/products/charmed-spark-3.4?ref=terraform-cc008"
 
-* Charmed Apache Spark 3.4.x
-  * [main `bundle.yaml`](https://github.com/canonical/spark-k8s-bundle/blob/main/releases/3.4/yaml/bundle.yaml.j2)
-  * [`overlays`](https://github.com/canonical/spark-k8s-bundle/blob/main/releases/3.4/yaml/overlays)
+  history_server_image       = "ghcr.io/canonical/charmed-spark:3.4-22.04_stable"
+  integration_hub_image      = "ghcr.io/canonical/spark-integration-hub:3-22.04_stable"
+  kyuubi_image               = "ghcr.io/canonical/charmed-spark-kyuubi:3.4-22.04_stable"
```
Batalex: **thought**: We should not have to explicitely pass the images **as a user**, I'll create a ticket to fix this. We should be able to do something nice with the juju_resource to automatically get the proper OCI for the revision while letting people override it if they want to

docs/how-to/deploy/spark.md
```diff
@@ -28,223 +28,208 @@ an S3-compatible object storage on MicroK8s (MinIO), EKS (AWS S3), or Azure obje
...
+| `admin_password`   | The password to set for the `admin` user that is used later to connect to Kyuubi   |
+| `tls_private_key` | The private key to be used for generating Kyuubi TLS certificates, provided as base64-encoded string |
+| `kyuubi_config.service-account` | The service account which is used by Kyuubi to run Spark jobs |
+| `storage_backend` | The object storage backend to be used. The backends `s3` and `azure_storage` are supported. |
+| `s3_access_key` | The S3 access key ID |
+| `s3_secret_key` | The S3 secret key |
+| `s3_config.endpoint` | The S3 endpoint |
+| `s3_config.region` | The S3 region |
+| `s3_config.bucket` | The name of the S3 bucket to be used |
+| `s3_config.path` | The path inside the S3 bucket to be used |
```
Batalex, line 104: **suggestion**: I believe it would easier and more maintainable to just link the module's readme here, since we have those variables and more automatically added to it

docs/how-to/deploy/spark.md
```diff
@@ -28,223 +28,208 @@ an S3-compatible object storage on MicroK8s (MinIO), EKS (AWS S3), or Azure obje
...
-To do so, first grant access to the secret for the Azure Storage Integrator charm
+resource "juju_model" "cos" {
+  name = "cos"
+}
 
-```shell
-juju grant-secret <secret_id> azure-storage
-```
+module "cos" {
+  source       = "git::https://github.com/canonical/observability-stack//terraform/cos-lite?ref=04ab6c618dbbec62292a052a61cdb402d80e5974"
```
Batalex: **nitpick**: Let's document what this specific commit sha mean (= track/2 branch from Apr 16, 2026)

**thought**: There were some commits since then, let's try to update sometime soon

docs/how-to/advanced-scheduling.md
```diff
@@ -228,6 +228,7 @@ juju deploy -m <charmed_spark_juju_model> kyuubi-k8s --trust --channel=3.5/edge
 A single Juju model can contain applications deployed over different architectures.
 
 ```{note}
+Make sure to use the correct track (eg, 3.4, 3.5, 4.0) for the version of Apache Spark that you'd like to use.
```
Batalex:
```suggestion
Make sure to use the correct track (e.g., 3.4, 3.5, 4.0) for the version of Apache Spark that you'd like to use.
```

Batalex (commented):
Yeaah, this new release looks super cool!

Shall we keep the reference to `terraform-cc008`, or already anticipate the next release tag?

docs/how-to/deploy/kyuubi.md
theoctober19th: Yeah, let's talk with Data team regarding this, once Mykola is back.

docs/how-to/deploy/spark.md
theoctober19th: Yeah -- that's what I was thinking as well.. At the moment, this thing is a mandatory variable and TF apply won't work without it. Maybe let's just create an issue out of this comment, so that we will come back to it sooner.

docs/how-to/deploy/spark.md
theoctober19th: I have indeed included the module's README a little down in the docs, I merely explained the most essential variables without which terraform apply would not work. Everything else that's not needed as "mandatory" is left here, and referred to the full README in `spark-k8s-bundle`.

theoctober19th:
> Shall we keep the reference to terraform-cc008, or already anticipate the next release tag?

@Batalex  I think I'm going to merge this PR **after** the `terraform-cc008` is merged to main. So I'll have a second pass coming at it to fix the links and references so we reference from `main`. For now, if I do so, the link checker would complain.

docs/how-to/deploy/spark.md
theoctober19th: According [to their docs](https://github.com/canonical/observability-stack/blob/track/2/terraform/cos-lite/README.md#provider--100--140) the correct hash to pin is `7448dadb996835c1c0ae1d79d2f435992652d410` so I will update this accordingly.

theoctober19th: As for updating this in the integration tests, those will be done in this PR: https://github.com/canonical/spark-k8s-bundle/pull/238 (since this PR contains doc change only and hence won't trigger integration tests).

docs/how-to/deploy/kyuubi.md
theoctober19th: I've updated the `text` with either (`shell`) or `bash` depending on what has been used in that specific file. Regarding replacing `bash` with `shell`, there are hundreds of places where both of these have been used interchangeably and I feel like updating all of them would be out of scope for this PR.

docs/how-to/deploy/kyuubi.md
theoctober19th: We'll talk this with the Data team and Mykola once he's back.

docs/how-to/deploy/spark.md
theoctober19th: [This commit](github.com/canonical/spark-k8s-bundle/pull/238/commits/8c9d0f305a7806083d4d5f0a5a2a2ceffd0bafae) from PR #238 should update this to the latest supported hash.

docs/how-to/deploy/spark.md
theoctober19th: https://warthogs.atlassian.net/browse/PRA-318

docs/how-to/deploy/spark.md
Batalex: I don't particulary mind having this table, though we could argue about what is essential or not here. `tls_private_key` and `kyuubi_config.service-account` could be omitted. Same with the images, once we fix this issue

Batalex (approved):

docs/how-to/deploy/spark.md
theoctober19th: I tried omitting TLS private key, but there's another problem that prevents us to not specify TLS private key. (see https://github.com/canonical/spark-k8s-bundle/issues/236).

For the images, yes you're right. We'll update the docs to remove the reference since after fixing that issue, those won't be necessary anymore.

docs/how-to/deploy/kyuubi.md
```diff
@@ -31,11 +31,12 @@ juju add-model kyuubi
 Deploy the Apache Kyuubi charm:
 
 ```bash
-juju deploy kyuubi-k8s --trust --channel=3.5/stable
+juju deploy kyuubi-k8s --trust --channel=3.4/stable
 ```
 
 ```{note}
-We also offer a  `3.4` track to run Apache Spark 3.4 workloads.
+We also offer a `3.5` and `4.0` track to run Apache Spark 3.4 and 4.0 workloads respectively.
```
izmalk: Suggestion: Fixing a typo with version nu,ber here.

```suggestion
We also offer a `3.5` and `4.0` track to run Apache Spark 3.5 and 4.0 workloads respectively.
```

docs/tutorial/1-environment-setup.md
izmalk: Suggestion: Shouldn't we update the version here too?


```suggestion
juju deploy s3-integrator --channel 2/stable
```

Make sure to implement neccessary relevant changes later in the tutorial.

izmalk (changes requested):
Looks good. A lot of neccessary fixes and updates. But I wonder whether it is fair to call this PR " Add docs for Spark 4.0". I don't see any specific documentation for 4.0 release. Not even release notes.
Is it OK if we rename this PR to something more specific and modest? Like: "[docs] Update docs for s3-integrator 2/stable and multi-track awareness". 

P.S. We have discussed with Paolo and Enrico the creation of a separate branch of docs for v.4.x. So we will have 3.x and 4.x. But to create 4.x we'll need at the very least some docs specific to 4.x. I can help with setting all up in RTD.

deusebio:
> P.S. We have discussed with Paolo and Enrico the creation of a separate branch of docs for v.4.x. So we will have 3.x and 4.x. But to create 4.x we'll need at the very least some docs specific to 4.x. I can help with setting all up in RTD.

Yes, it is probably best to have two branches for the `3/` and `4/` documentation, given the ability of RTD to manage multiple branches neatly. However, given that we only have `main` here, I'm wondering whether the way to do this would be to migrate the documentation to a proper repository. Maybe this is worth discussing in a daily

theoctober19th:
Hi @izmalk, I've addressed your comments in the later commits, and also have updated the PR title. Can you please have another look at the PR?

@deusebio Regarding creating multiple tracks for the docs -- I'd rather think it would be better to have a dedicated ticket for the docs migration and not block Spark 3.4/3.4/4.0 release due to this. 

While having multiple doc branches is a nice thing to have, I think at this point this may even feel like an overkill (because the only thing that we have different across various Spark versions is the image tags and the track from where the charms are to be deployed). However, this is also a tradeoff between effort vs value -- and I agree 100% with going forward with having multiple docs as long as the value makes up for the effort. In either case, I think the way forward would be to handle this in dedicated ticket / PR because we would most likely move the docs to a separate repo.

docs/tutorial/1-environment-setup.md
theoctober19th: Updated this in [a220af9](https://github.com/canonical/spark-k8s-bundle/pull/237/commits/a220af970465841817ac17e924d61ded6ae4783e).

deusebio (changes requested):
The changes look fine and sane. The only thing I'd like to change is to refer to 3.5 as default. This is probably going to be our LTS on the `3/` branch, so I'd rather start to use that one as a general guidance when deploying instead of going to 3.4 and then saying there is also 3.5

theoctober19th:
@deusebio I've addressed your comment on using Spark 3.5 by default in commit [1a14e5c](https://github.com/canonical/spark-k8s-bundle/pull/237/commits/1a14e5ccdeb4124a292955acf7579ba10714e5b8).

deusebio (approved):
Looks good! Thank you for this, Bikalpa!

docs/how-to/deploy/kyuubi.md
```diff
@@ -96,23 +105,27 @@ After that, `kyuubi-k8s` app becomes blocked with the `Missing authentication da
...
 
-Create a Juju secret containing the Azure Storage account secret key, and grant the charm access to this secret.
+Create a Juju secret containing the Azure Storage account secret key, and grant the charm access to this secret. 
+Make note of the output of the `juju add-secret` command; this is the secret URI for the added secret that we will need
+in the next step.
 
-```text
+```bash
 juju add-secret azure-storage-secret secret-key=XXXXXXX
+# secret:jem6a4josup6rui0g2q0  <-- make note of this secret URI
```
izmalk: Suggestion: I really don't like such usage of comments in code blocks.
The instructions here can be a bit confusing. I would advise to put the instructions in a generic text block with example of an output to be a dedicatet code block with proper introduction.

docs/tutorial/1-environment-setup.md
```diff
@@ -483,17 +488,42 @@ Deploy the Integration Hub charm and an `s3-integrator` charm to supply it with
...
-The `s3-integrator` will remain in `blocked` state until S3 credentials are provided. Set the credentials first, then integrate:
+The `s3-integrator` will remain in `blocked` state until S3 credentials are provided. 
+
+Create a Juju secret that contains the S3 access key and secret key, and grant the secret to the `s3-integrator` app.
+Make note of the output of the `juju add-secret` command; this is the secret URI for the added secret that we will need
+in the next step.
+
+```bash
+juju add-secret s3-creds access-key=<ACCESS-KEY> secret-key=<SECRET-KEY>
+# secret:jem6a4josup6rui0g2q0  <-- make note of this secret URI
```
izmalk: Suggestion: The same formatting issue with the instructions in comments of a code block

docs/tutorial/1-environment-setup.md
```diff
@@ -483,17 +488,42 @@ Deploy the Integration Hub charm and an `s3-integrator` charm to supply it with
...
 <!-- test:await-idle --timeout 600 --allow-blocked s3-integrator -->
 
-The `s3-integrator` will remain in `blocked` state until S3 credentials are provided. Set the credentials first, then integrate:
+The `s3-integrator` will remain in `blocked` state until S3 credentials are provided. 
+
+Create a Juju secret that contains the S3 access key and secret key, and grant the secret to the `s3-integrator` app.
+Make note of the output of the `juju add-secret` command; this is the secret URI for the added secret that we will need
+in the next step.
+
+```bash
```
izmalk: Note: This is a bash type code block. Which means it won't be executed by the tutorial testing. If we want to actually add this secret, we need to either make the block to be shell type, or add it with an explicit command in comments.

docs/tutorial/4-history-server.md
```diff
@@ -59,39 +59,60 @@ during the environment setup step.
...
 
-<!-- test:await-idle --timeout 600 --allow-blocked s3-integrator,spark-history-server-k8s -->
+Create a Juju secret that contains the S3 access key and secret key, and grant the secret to the `s3-integrator` app.
+Make note of the output of the `juju add-secret` command; this is the secret URI for the added secret that we will need
+in the next step.
 
-Wait a minute for the `s3-integrator` charm to deploy correctly before finishing the setup:
+```bash
+juju add-secret s3-creds access-key=<ACCESS-KEY> secret-key=<SECRET-KEY>
+# secret:jem6a4josup6rui0g2q0  <-- make note of this secret URI
```
izmalk: Suggestion: The same formatting issue with the instructions in comments of a code block

izmalk (changes requested):
Great work.
I see that you have added changes to the tutorial section. Have you tried running the tutorial test? We update some versions and add new commands, I'd like to make sure we tested it before meging.

Lastly, I've noticed you used a pattern where you provide important instructions inside comments of code blocks. I would really want us to avoid doing that if possible. All important instructions should be rendered outside code blocks as generic text. Let me know if you feel strongly about it, we can discuss alternatives.

docs/tutorial/1-environment-setup.md
```diff
@@ -483,17 +488,42 @@ Deploy the Integration Hub charm and an `s3-integrator` charm to supply it with
...
+juju add-secret s3-creds access-key=<ACCESS-KEY> secret-key=<SECRET-KEY>
+# secret:jem6a4josup6rui0g2q0  <-- make note of this secret URI
+
+juju grant-secret s3-creds s3-integrator
+```
+
+Configure S3 integrator to use your S3 object storage by setting up the credentials (make sure to pass the secret URI from previous step in place of `$SECRET_URI`):
+
+<!-- test:run
+export SECRET_URI=$(juju show-secret s3creds | yq 'keys | .[0]')
```
izmalk: Chore: I see an error with yq being unavailable in the test environment.

```
/spark-k8s-bundle/python/tests/tutorial/1-environment-setup.sh: line 108: yq: command not found
```

That error leades to more problems later as we fail to properly setup the storage.

izmalk (changes requested):
I've tested the tutorial in it's current state and it shows 3 out of 7 steps green.
Most of the problems are caused by the failed S3 setup due to missing `yq` in the testing environment.
Here are the full logs - https://gist.github.com/izmalk/9d2b009f13346868bc92ef82261c872d
I've also added a PR to add CI workflow - https://github.com/canonical/spark-k8s-bundle/pull/239

docs/tutorial/4-history-server.md
theoctober19th: Fixed this in [604d5cc](https://github.com/canonical/spark-k8s-bundle/pull/237/commits/604d5cce50fe4a8efbf96bc818e917240b2c4a1d)

docs/tutorial/1-environment-setup.md
theoctober19th: Fixed this in [604d5cc](https://github.com/canonical/spark-k8s-bundle/pull/237/commits/604d5cce50fe4a8efbf96bc818e917240b2c4a1d).

docs/tutorial/1-environment-setup.md
theoctober19th: Fixed in [604d5cc](https://github.com/canonical/spark-k8s-bundle/pull/237/commits/604d5cce50fe4a8efbf96bc818e917240b2c4a1d).

docs/how-to/deploy/kyuubi.md
theoctober19th: Fixed in [604d5cc](https://github.com/canonical/spark-k8s-bundle/pull/237/commits/604d5cce50fe4a8efbf96bc818e917240b2c4a1d).

docs/tutorial/1-environment-setup.md
theoctober19th: I've fixed this particular line in the docs test, however I'm still not able to reliably run tutorial tests locally. I'll ask for re-review once I have them passing locally.

.github/workflows/tutorial-tests.yaml
```diff
@@ -0,0 +1,67 @@
...
+name: End-to-end Tutorial Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    paths:
+      - 'docs/**'         # Run on changes to the docs directory
+      - 'spread.yaml'     # Run when spread.yaml is modified (e.g., changes to the bootstrapping configuration)
+      - 'python/tox.ini'  # Run when the tox.ini file is modified (e.g., changes to the tutorial test environment)
```
izmalk, line 13: Question: Are you sure this is neccessary?
I feel like we should not track the changes here for this particular test. Both files can change for something else and almost never need to be changed for tutorial test changes.

docs/how-to/use-k8s-pods.md
```diff
@@ -66,15 +71,22 @@ python3 -m spark8t.cli.service_account_registry create --username spark
...
+spark-client.spark-submit --username spark --class org.apache.spark.examples.SparkPi local:///opt/spark/examples/jars/spark-examples_2.12-3.5.8.jar 100
 ```
 
+```{note}
+Please make sure to use the correct JAR for the version of Apache Spark and Scala that you'd like to use.
+For instance, if you want to use Apache Spark 4.0.2 (which natively supports Scala 2.13), 
+you should use the JAR `spark-examples_2.13-4.0.2.jar` instead.
+```
+
+
```
izmalk:
```suggestion
```

izmalk (approved):
LGTM.
Just two non-blocking suggestions below. Nothing critical.

.github/workflows/tutorial-tests.yaml
theoctober19th: This is a good thing to have. If someone accidentally changes something in `tox -e tutorial` or the spread environment, and if the tests do not run, we would never know the changes broke them. On the other hand i don't think these tests would make the PR slower anyway, because if you touch those files, around 40+ tests in the CI will need to run and signal green for you to be able to merge the PR.. and this workflow will run in parallel to them.

PR #239, opened 2026-06-23
docs: Add GH workflow for tutorial tests

izmalk:
## Description

This is to make tutorial tests to run:

- When triggered by other workflows (for example, release or risk promotion)
- When manually triggered by a button on GH
- Once in a while on schedule (monthly in this particular case, but we can easily adjust) - just to detect an unexpected regression

Beased on the [Kafka GH workflow](https://github.com/canonical/kafka-operator/blob/main/.github/workflows/tutorial-tests.yaml).

## Checklist

### DIFF
--- a/.github/workflows/tutorial-tests.yaml
+++ b/.github/workflows/tutorial-tests.yaml
@@ -1,0 +1,55 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: Tutorial tests
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  schedule:
+    - cron: '0 3 2 * *'  # 2nd of every month at 03:00 UTC
+
+jobs:
+  tutorial-test:
+    name: Tutorial end-to-end
+    runs-on: [self-hosted, linux, AMD64, X64, xlarge, noble]
+    timeout-minutes: 180
+    steps:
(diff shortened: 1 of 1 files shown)


theoctober19th (commented):
Code-wise everything looks good. It would be worth however to run this test on every PR, what do you think on this? Once add the `pull_request` trigger, we can also right away test whether it passes or not in CI in this very PR.

izmalk:
> It would be worth however to run this test on every PR, what do you think on this?

Well, I don't mind, but at the very least, let's limit it to only PRs that have changes to tutorial files.

> Once add the pull_request trigger, we can also right away test whether it passes or not in CI in this very PR.

Really? I thought it doesn't work until we merge it into main.

theoctober19th:
> Well, I don't mind, but at the very least, let's limit it to only PRs that have changes to tutorial files.

Yes, that's fine. We can do this only for changes in tutorial files.

> Really? I thought it doesn't work until we merge it into main.

It will work on all PR, also on feature branches even before we merge to main.

theoctober19th:
Hi @izmalk, I think this workflow would be very helpful in the other PR #237 itself, since we can then assure that it is working. So, I'm gonna merge this PR to `spark-4-docs` branch first, and from there we can take it to the base branch.

PR #240, opened 2026-06-24
[PRA-318] Implement automated OCI getter for our products

Batalex:
This PR fixes the issue of making the OCI resources for our products mandatory. 

In addition, it adds a new setup that I find quite neat. Thanks to the "juju_charm" resource, we have some automation that:
-  picks latest charm revision on the track, with the latest oci resource
- if we just override the charm revision, then it automatically picks the corresponding oci resource
- allows to override charm revision and oci resource, or just the oci resource

Here are some results from my own tests:

- Without variables, we deploy the history server revision 98 and OCI rev 24
- When we pass `history_server_image="ghcr.io/canonical/charmed-spark@sha256:f7f387a76ba2f3b6cfc34a35f36c0f4ae85bfa78c8c9d4bfa06ba86401b95d70"`  we get the following plan
```
 # module.spark.module.spark_core.juju_application.history_server will be updated in-place
  ~ resource "juju_application" "history_server" {
        id          = "55cdab9f-9342-47a1-88ea-698499732018:history-server"
      ~ machines    = [] -> (known after apply)
        name        = "history-server"
      ~ resources   = {
          ~ "spark-history-server-image" = "24" -> "ghcr.io/canonical/charmed-spark@sha256:f7f387a76ba2f3b6cfc34a35f36c0f4ae85bfa78c8c9d4bfa06ba86401b95d70"
        }
      + storage     = (known after apply)
        # (6 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```
- When we pass `history_server_revision=117`, we get
```
 # module.spark.module.spark_core.juju_application.history_server will be updated in-place
  ~ resource "juju_application" "history_server" {
        id          = "55cdab9f-9342-47a1-88ea-698499732018:history-server"
      ~ machines    = [] -> (known after apply)
        name        = "history-server"
      ~ resources   = {
          ~ "spark-history-server-image" = "24" -> "29"
        }
      + storage     = (known after apply)
        # (6 unchanged attributes hidden)

      ~ charm {
            name     = "spark-history-server-k8s"
          ~ revision = 98 -> 117
            # (2 unchanged attributes hidden)
        }
    }
```
So we appropriately pick the proper oci resource
- When we pass both, we get
```
 # module.spark.module.spark_core.juju_application.history_server will be updated in-place
  ~ resource "juju_application" "history_server" {
        id          = "55cdab9f-9342-47a1-88ea-698499732018:history-server"
      ~ machines    = [] -> (known after apply)
        name        = "history-server"
      ~ resources   = {
          ~ "spark-history-server-image" = "24" -> "ghcr.io/canonical/charmed-spark@sha256:f7f387a76ba2f3b6cfc34a35f36c0f4ae85bfa78c8c9d4bfa06ba86401b95d70"
        }
      + storage     = (known after apply)
        # (6 unchanged attributes hidden)

      ~ charm {
            name     = "spark-history-server-k8s"
          ~ revision = 98 -> 117
            # (2 unchanged attributes hidden)
        }
    }
```

Perhaps something we can share with folks working on charm modules as well?

### DIFF
--- a/terraform/charms/azure-storage-integrator/README.md
+++ b/terraform/charms/azure-storage-integrator/README.md
@@ -16,7 +16,7 @@ To be contributed upstream.
 
 | Name | Version |
 | ---- | ------- |
-| <a name="provider_juju"></a> [juju](#provider\_juju) | 1.3.1 |
+| <a name="provider_juju"></a> [juju](#provider\_juju) | >=1.0.0 |
 
 ### Modules
 
--- a/terraform/charms/s3-integrator/README.md
+++ b/terraform/charms/s3-integrator/README.md
@@ -16,7 +16,7 @@ To be contributed upstream.
 
 | Name | Version |
 | ---- | ------- |
-| <a name="provider_juju"></a> [juju](#provider\_juju) | 1.3.1 |
+| <a name="provider_juju"></a> [juju](#provider\_juju) | >=1.0.0 |
 
 ### Modules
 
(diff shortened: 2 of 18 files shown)


terraform/charms/azure-storage-integrator/README.md
```diff
@@ -16,7 +16,7 @@ To be contributed upstream.
 
 | Name | Version |
 | ---- | ------- |
-| <a name="provider_juju"></a> [juju](#provider\_juju) | 1.3.1 |
+| <a name="provider_juju"></a> [juju](#provider\_juju) | >=1.0.0 |
```
theoctober19th, line 19: this is a nice improvement.. rather than pinning on one specific version.

deusebio (approved):
While reading the PR, I initially had some seconds thoughts about the fact that this may trigger some updates when people do not pin either revisions or images, and therefore when they do apply on a living deployment, if a new revision is published, that may be consumed and therefore an upgrade trigger although nothing has changed in the module. 

But this is probably expected behaviour when you don't pin. Also, people should in general pin and also do a plan before applying and checking what would happen. If something would update, and they don't want that, they should THEN pin. Right now we generally don't support upgrades through TF operations, so this part is also a gray area where further experimentation may be required, but I'm happy to give this a go. Indeed, the UX that we get is quite natural and neat.

theoctober19th (approved):
LGTM, thanks for the PR, Alex!

I agree with what @deusebio  mentioned in his review comment above -- on the fact that if the users do not pin the images (which would be the default behavior), and we rely on Juju to correctly choose the images for the tracks we specify in the TF modules, the correct behavior on `terraform apply` would indeed be to refresh the charms to newer revisions if the images have evolved forward.

I would honestly also would have lived with not having this change in the first place -- only if it wouldn't be so cumbersome for the users to go find the appropriate images to be supplied to TFVars, or else the `terraform apply` would not work! As such, it is easy for the users to get confused on where to get these images from or annoyed at the least. With this change, we eliminate the steep curve for the users who quickly want to try our TF modules, while still providing for the feature for advanced users to keep the images pinned if they liked.

TLDR; I'm very happy with the change. 

@Batalex I'm also going to merge this PR (since with my review there'll be 2 required approvals), as this will unblock me to work on rebasing `terraform-cc008` early morning tomorrow (since you're already past your EOD)

PR #241, opened 2026-06-24
[PRA-312] Split TLS private key and admin password secrets

### LINKED ISSUE #236: The `tls-client-private-key` is an optional config option for `kyuubi-k8s` but the bundle makes it necessary

Batalex:
This PR fixes #236 by splitting the existing secret in two

### DIFF
--- a/terraform/products/charmed-spark-3.5/main.tf
+++ b/terraform/products/charmed-spark-3.5/main.tf
@@ -168,7 +168,7 @@ resource "juju_secret" "private_key_secret" {
     private-key = var.tls_private_key
   }
 
-  info = "This secret contains the TLS private key."
+  info = "This secret contains the base64 encoded TLS private key."
 }
 
 resource "juju_secret" "system_users_secret" {
--- a/terraform/products/charmed-spark-3.5/main.tf
+++ b/terraform/products/charmed-spark-3.5/main.tf
@@ -159,20 +159,28 @@ resource "juju_access_secret" "s3_secret_access" {
   secret_id = juju_secret.s3_secret[0].secret_id
 }
 
-resource "juju_secret" "system_users_and_private_key_secret" {
+resource "juju_secret" "private_key_secret" {
   depends_on = [juju_model.spark]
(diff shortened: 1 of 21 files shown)


terraform/products/charmed-spark-3.5/main.tf
```diff
@@ -159,20 +159,28 @@ resource "juju_access_secret" "s3_secret_access" {
...
-      private-key = var.tls_private_key
-    }
-  )
-  info = "This secret contains password for admin user and the TLS private key."
+  name       = "private_key_secret"
+  value = {
+    private-key = var.tls_private_key
+  }
+
+  info = "This secret contains the TLS private key."
```
theoctober19th: nit: worth mentioning that this should contain **base64-encoded** string of the actual key.

theoctober19th (approved):
LGTM. Thanks for the quick fix.

terraform/products/charmed-spark-3.5/main.tf
Batalex: Specified in 2668e957bbbaa2b9cff2ed982ac4553fba77f805

deusebio (approved):
Looks a sensible change/fix!

theoctober19th:
Merging this PR as I need this to work on rebasing `terraform-cc008` and there are 2 approvals already.

PR #246, opened 2026-06-27
[PRA-330] Bump postgresql charm to 16/stable on track/4.0

theoctober19th:
## Description

Update the postgresql charm (metastore and kyuubi-users) to `16/stable` and update docs.

## Checklist

### DIFF
--- a/docs/how-to/apache-kyuubi/external-metastore.md
+++ b/docs/how-to/apache-kyuubi/external-metastore.md
@@ -19,7 +19,7 @@ The Charmed Apache Kyuubi K8s charm provides a `metastore-db` integration throug
 To use it, deploy a Charmed PostgreSQL K8s charm:
 
 ```shell
-juju deploy postgresql-k8s metastore --channel 14/stable --trust
+juju deploy postgresql-k8s metastore --channel 16/stable --trust
 ```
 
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-uat.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         spark-version:
-          - "3.4.4"
+          - "3.5.8"
```
Batalex: **question**: Is 3.5 the default track from now on? I don't have any big concerns; I just don't remember us discussing this in dailies

Batalex (approved):
LGTM

.github/workflows/ci-uat.yaml
theoctober19th: I believe we want to make `3.5` the default, since it is LTS version and also we have our docs edited everywhere to use 3.5 by default.

deusebio (changes requested):
I believe we discussed to change this only for Spark 4, not for spark 3.x. This is a major upgrade and - given there is no upgrade path between 14 and 16 - I would only do this for Spark 4. Let's discuss at daily if needed

deusebio (approved):

PR #248, opened 2026-06-29
[DO NOT MERGE][PRA-324] Split bundle by track (4.0)

Batalex:
This PR showcases the new split-by-track structure we want to migrate this repository to.

This PR will be closed once approved, and instead the `track/4.0` branch will be squashed and live on.

## Changes

- The python part is basically untouched, except for the removal of the `spark-version` pytest arg. We now have a simple constant to update in `conftest.py`
- Removed the unneeded parts of the tf modules, but kept the same structure, so that it's easier to potentially merge two branches in one folder
- Dropped the `spark-version` dimension from the CI, which means that we now run the UAT on every track, but each PR will see a significant reduction in jobs


theoctober19th (approved):
The changes are exactly what I had expected, so LGTM.

One thing that I was also talking earlier with @Batalex was the structure we would want to put our TF modules into now. I see that we still retain `terraform/products/charmed-spark-4.0` structure. While I see that this will leave room for potential merging of multlple branches to a single folder later in the future, on the other hand the terraform modules feel like they are unnecessarily nested up inside. (As for example, I have seen many charms and solutions have their TF modules root directly on `terraform`. I'd also like to hear @deusebio 's take on this.

.github/actions/run-test/action.yaml
```diff
@@ -18,11 +18,6 @@ inputs:
     default: 1.32
     required: false
     type: string
-  spark-version:
```
deusebio, line 21: **non-blocking** We could have left this to just allow to run different patch version. But this is not blocking. If useful, we can re-introduce this at a later stage

deusebio (approved):
Looks good! 

Regarding @theoctober19th comment, I would probably just rename the product removing the version, which is included now in the branch. Maybe we could also renamed the product `charmed-kyuubi` given the main service provided by this solution is actually Kyuubi. 

So I would leave the structure `terraform/products/<name>`, such that we could have multiple products eventually. But please no version there as it seems a bit redundant. Maybe check with @grobbie what the name of the solution could be. So far, I'd vote for `charmed-kyuubi`

Batalex:
We are good to go, let's close this and make `track/4.0` a proper development branch

PR #252, opened 2026-06-30
[PRA-324] Adapt docs content to 3.5 track

Batalex:
Our docs were using 3.4 as the reference in a few places, so I changed that.
The contributing guide was also outdated

I also reverted the terraform provider pin since we no longer need it


deusebio (approved):
Thanks @Batalex  for the chore work here!

welpaolo (approved):
LGTM

theoctober19th (approved):

PR #254, opened 2026-07-01
[PRA-324] Adapt docs content to 4.0 track

Batalex:
Similar to #252.

Notably, we run the end-to-end tutorial using Spark 4


welpaolo (approved):
LGTM

docs/how-to/apache-kyuubi/gpu.md
```diff
@@ -13,7 +13,7 @@ This makes it possible to run queries with hardware acceleration on NVIDIA GPUs.
 ## Prerequisites
 
 - Kubernetes cluster is up and running with NVIDIA GPU support
-- Charmed Apache Kyuubi, revision `121` (Apache Spark 3.5), `122` (Apache Spark 3.4) or higher
+- Charmed Apache Kyuubi, revision `TBD` (Apache Spark 4.0), `121` (Apache Spark 3.5), `122` (Apache Spark 3.4) or higher
```
deusebio, line 16: **to-do** can we please create a ticket in the backlog to make sure we update this reference? I'm just afraid we will forget and having a TBD is not that nice in the docs. We could also start to put what we know it is the current revision, such that if we forget, it will just look outdated, rather than TBD

deusebio (approved):
Nice work! Thanks Alex! Just a small comment, but once addressed, feel free to merge straight away. I don't need to re-review

docs/how-to/apache-kyuubi/gpu.md
Batalex: Sure, created #290 for it!

PR #280, opened 2026-07-03
[PRA-323] chore: docs migration to canonical.com

deusebio:
## Description

Changes required to do the canonical.com migration

## Checklist

### DIFF
--- a/docs/_static/js/overwrite_links.js
+++ b/docs/_static/js/overwrite_links.js
@@ -1,0 +1,52 @@
+// Replaces rtd-address with new-address in links
+
+const rtd_address = 'canonical-charmed-spark.readthedocs-hosted.com';
+const new_address = 'canonical.com/data/spark/docs';
+
+function escapeRegExp(value) {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function overwriteMatchingAnchorUrls(container) {
+  if (!container) return;
+
+  const rtd_addressRegex = new RegExp(escapeRegExp(rtd_address), 'g');
+  container.querySelectorAll('a[href], link[href]').forEach((anchor) => {
+    anchor.href = anchor.href.replace(rtd_addressRegex, new_address);
+  });
+}
+
+function patchFlyout() {
+  const rtdFlyout = document.querySelector('readthedocs-flyout');
+  if (!rtdFlyout) return false;
+
+  overwriteMatchingAnchorUrls(rtdFlyout);
+  overwriteMatchingAnchorUrls(rtdFlyout.shadowRoot);
+
+  rtdFlyout.addEventListener('click', () => {
+    overwriteMatchingAnchorUrls(rtdFlyout);
+    overwriteMatchingAnchorUrls(rtdFlyout.shadowRoot);
+  });
+
+  return true;
+}
+
+function init() {
+  overwriteMatchingAnchorUrls(document.querySelector('header'));
+
+  if (patchFlyout()) return;
+
+  const observer = new MutationObserver(() => {
+    if (patchFlyout()) {
+      observer.disconnect();
+    }
+  });
+
+  observer.observe(document.body, { childList: true, subtree: true });
+}
+
+if (document.body) {
+  init();
+} else {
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -169,24 +159,31 @@
 # TODO: If your documentation is hosted on https://docs.ubuntu.com/,
 #       uncomment and update as needed.
 
-# slug = 'charmed-spark'
+slug = 'data/spark/docs'
```
deusebio, line 160: @grobbie is this the slug that we want to have for Spark? I'm wondering whether `data` should be there

izmalk: The current approach is to have it as `<product-page>/docs`.
Is the product page still http://canonical.com/data/spark/ ?

docs/conf.py
```diff
@@ -169,24 +159,31 @@
...
 
 #######################
 # Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
 #######################
 
 # Base URL of RTD hosted project
 
-html_baseurl = 'https://canonical-charmed-spark.readthedocs-hosted.com/'
+html_baseurl = f"https://canonical.com/{slug}/{os.environ.get('READTHEDOCS_VERSION', 'local')}"
```
izmalk, line 168: That is a bit unexpected use of the slug for me, but I don't see any problem with that.

docs/conf.py
```diff
@@ -316,6 +313,7 @@
 
 html_js_files = [
 	'js/bundle.js',
+        'js/overwrite_links.js',
```
izmalk: Question: I don't see that file being added in this PR. I think we need to place the file in the correct directory to link to it.

Additionally, indentation is different from the bundle.js line for some reason, but it's not critical.

izmalk (changes requested):
I've noticed we are lacking the JS file - https://documentation.ubuntu.com/rtd-proxy/how-to/migrate/#overwrite-links-js
We referencing to it, but there is no such file in the repo yet.

docs/conf.py
deusebio: ops, had been left out from the commit! Nice catch!

docs/conf.py
deusebio: Yes, it seemed a better way to do this rather than copy and paste, since the variable was already defined.

docs/conf.py
deusebio: yes, it is. I also checked with Rob and he is fine with this

docs/conf.py
izmalk: Do you think the indentation of the line is correct?

docs/_static/js/overwrite_links.js
```diff
@@ -0,0 +1,52 @@
+// Replaces rtd-address with new-address in links
+
+const rtd_address = 'canonical-charmed-spark.readthedocs-hosted.com';
```
izmalk: Todo: For the whole thing to work, this needs to be the new proxy project's address, IMHO. `canonical-charmed-spark-proxy.readthedocs-hosted.com`

izmalk (changes requested):
I think there is a problem with the RTD address to redirect from

docs/_static/js/overwrite_links.js
deusebio: should be fixed

deusebio:
@izmalk It seems that the documentation is working on staging, see https://staging.canonical.com/data/spark/docs

Once you approve this PR, I'll request your reviews also on the companions PRs:
* #281 
* #282

izmalk:
> @izmalk It seems that the documentation is working on staging, see https://staging.canonical.com/data/spark/docs

Good work, I see it is working on staging. However, I see a problem with the JS redirects. Did you check the installation as per this section: https://documentation.ubuntu.com/rtd-proxy/how-to/migrate/#url-rewriting ?

The links in the RTD overlay dropdown are not correct (should be canonical.com). In fact, they are not even from the proxy-project, but from the original project, which surprises me.

Batalex (approved):
AFAICT it follows the migration guide, LGTM

welpaolo (approved):

izmalk (approved):
LGTM

PR #282, opened 2026-07-03
[PRA-323] chore: docs migration to canonical.com (#280)

deusebio:
## Description

Changes to allow for the migration to canonical.com

## Checklist

### DIFF
--- a/docs/_static/js/overwrite_links.js
+++ b/docs/_static/js/overwrite_links.js
@@ -1,0 +1,52 @@
+// Replaces rtd-address with new-address in links
+
+const rtd_address = 'canonical-charmed-spark.readthedocs-hosted.com';
+const new_address = 'canonical.com/data/spark/docs';
+
+function escapeRegExp(value) {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function overwriteMatchingAnchorUrls(container) {
+  if (!container) return;
+
+  const rtd_addressRegex = new RegExp(escapeRegExp(rtd_address), 'g');
+  container.querySelectorAll('a[href], link[href]').forEach((anchor) => {
+    anchor.href = anchor.href.replace(rtd_addressRegex, new_address);
+  });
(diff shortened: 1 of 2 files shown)


docs/_static/js/overwrite_links.js
```diff
@@ -0,0 +1,52 @@
+// Replaces rtd-address with new-address in links
+
+const rtd_address = 'canonical-charmed-spark-proxy.readthedocs-hosted.com';
```
theoctober19th, line 3: I didn't understand why there's a `proxy` in the address here, and also could not access this link anymore from the browser.

Also, wouldn't it be better if we update the doc links permanently than rely on this Javascript to do it on-the-fly?

deusebio: This is according to the guidance provided here: https://documentation.ubuntu.com/rtd-proxy/how-to/migrate/. 

Also @izmalk has been reviewing the "main" PR on 3.5, see here: https://github.com/canonical/spark-k8s-bundle/pull/280

Let's comment/merge that one, and then I'll backport things to this one

Batalex (approved):

docs/_static/js/overwrite_links.js
izmalk: It's all according to the plan.
Unfortunately, we can't natively publish our docs on canonical.com via RTD, as this is the most important domain and we manage it ourselves. So this is a part of the complex URL migration solution we have for our docs right now.
The proxy address is the address published by the new RTD project. The new RTD project is private so people won't go there directly. Instead, we have our RTD proxy serving this same website throught the canonical.com/data/spark/docs address.
But we still need to adjust links on the page, and especially the version selector overlay, to lead to canonical.com and not the original readthedocs-hosted.com. We do that we this JS.

izmalk (approved):
LGTM

welpaolo (approved):

theoctober19th (approved):

